PR #50, opened 2015-03-14
Add a sync function 'createPoolSync'

karboom:
Sometimes people don't want wrap their app in the createPool's callback, so I write a sync version

### DIFF
--- a/src/njs/src/njsOracle.cpp
+++ b/src/njs/src/njsOracle.cpp
@@ -90,6 +90,7 @@ void Oracledb::Init(Handle<Object> target)
 
   NODE_SET_PROTOTYPE_METHOD(oracledbTemplate_s, "getConnection", GetConnection);
   NODE_SET_PROTOTYPE_METHOD(oracledbTemplate_s, "createPool", CreatePool);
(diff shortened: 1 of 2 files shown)


cjbj:
Before merging, this is worthy of some more discussion about timeouts vs sync vs async calls.  I'm glad you created the PR so it is in concrete form so the direction that users want can be seen.

To help your ideas stand out, can you update the PR without all the whitespace changes? (I will hassle the maintainers about stripping whitespace so diffs are easier in future)

Can you review CONTRIBUTING.md?

karboom:
I have solved crlf problem, how can I get OCA?

cjbj:
@karboom Details on OCA are in https://github.com/oracle/node-oracledb/blob/master/CONTRIBUTING.md

We want to revisit the wider topic of how calls can have timeouts, and review async vs sync calls.  Only when this overall review is done will it be possible to see how your patch fits in.  I don't have an ETA on this.  I'm not saying your work won't be used. I just can't tell.  At the very least, the fact the feature is important to you will guide us.

cjbj:
Closing this since the OCA paperwork wasn't submitted.

PR #156, opened 2015-08-07
fixed crash when result sets have unsupported types

bjouhier:
This one was trivial. You can discard if you already fixed it.

Signed-off-by: Bruno Jouhier <bruno.jouhier@sage.com>


cjbj:
Thanks @bjouhier.  Your investigation is much appreciate. We do indeed have a fix ready. It is similar to yours but also sets getRowsBaton->error = ebaton->error

PR #167, opened 2015-08-17
Fixed outFormat: oracledb.OBJECT example output

nethoncho:
The example output for oracledb.OBJECT was the same as oracledb.ARRAY

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -1941,8 +1941,8 @@ connection.execute("SELECT first_name, salary, hire_date "
 If run with Oracle's sample HR schema, the output is:
 
 ```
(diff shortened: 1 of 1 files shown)


nethoncho:
Just noticed the value for SALARY on row 0 should be 12000 instead of 12008

cjbj:
@nethoncho Sorry to be formal, but can you sign the OCA? See https://github.com/oracle/node-oracledb/blob/master/CONTRIBUTING.md

cjbj:
@nethoncho I've reworked the examples in this section and fixed the output along the way. 
Thanks for reporting the problem.

Issue #189, opened 2015-09-03
copy the installation from one computer to another ?


rogergl:
Is it possible to copy the oracledb installation from one computer to another ? I tried to copy the complete node_modules folder, containing the oracledb installation, and the instant client folder from the development server to the production server. I installed the Visual Studio 2013 redistributables but I get the following error  as soon as I try to require the oracledb module:

C:\node_modules\oracledb\lib\oracledb.js:28
    throw err;
          ^
Error: The specified module could not be found.
C:\node_modules\oracledb\build\Release\oracledb.node
    at Error (native)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (C:\node_modules\oracledb\lib\oracledb.js:23:15)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)

cjbj:
I suspect it's possible in some circumstances, which is why things like node-pre-gyp exist, but it has not been investigated by us.  Check you have set PATH to the location of Instant Client

pvenkatraman:
Have you set the required environment variables.  Check %NODE_PATH% particularly.

rogergl:
Yes, I have set the PATH. But it still does not work.

bchr02:
Okay, so I tried doing the same exact thing as you and had the same exact problem. The solution was to install [Microsoft Visual C++ 2010 Redistributable](http://www.microsoft.com/en-us/download/details.aspx?id=26999).

Here is how I figured this out. But first, I want to mention my testing environment (in case anyone is curious). 
* Windows 7 64bit
* 64bit Node 0.12.7
* Visual Studio Community 2013 with Update 5 (which also installed Visual C++ 2012 Redistributable)

My initial reaction was by moving a node app which has a compiled copy of node-oracledb, it should work simply by copying it to another 64bit Windows computer with Visual C++ 2012 Redistributable. But this was not the case and instead I got the same error msg:
> Error: The specified module could not be found.

[...]

PR #191, opened 2015-09-04
better explained Visual C++ Redistributable requirement

bchr02:
node-oracledb only requires Visual C++ Redistributable when a compiler is not installed. Also, the version that seems to be required is 2010 (see https://github.com/oracle/node-oracledb/issues/189#issuecomment-137874533) and not the version that gets installed automatically with the Visual Studio Compiler.


cjbj:
Thank you @bchr02.  I'll let @pvenkatraman review this

bchr02:
https://github.com/oracle/node-oracledb/commit/5283c3be1d18635ef6e12f794a7635757bba0681 takes care of this. Thanks @cjbj

Issue #208, opened 2015-09-15
LOB stream emits close before end


bjouhier:
I'm getting a premature `close` event when reading a LOB. I'm expecting to get one or more `data` events followed by an `end` event and then a `close` event. 

Instead I'm getting the `data` events followed by a `close` event.

cjbj:
Thanks @bjouhier

cjbj:
Fixed in node-oracledb 1.2.  
Release notes are at: https://blogs.oracle.com/opal/entry/node_oracledb_1_2_0

PR #209, opened 2015-09-15
fixed oracle/node-oracledb#208 - emit end event before close

bjouhier:
This is a simple fix to #208. The `null` value must be pushed before emitting the `close` event and the `close` event must postponed with a `process.nextTick` call.


cjbj:
This was merged to node-oracledb 1.2.  Thanks @bjouhier

Issue #210, opened 2015-09-15
driver crashes if reading several rows with LOBs.


bjouhier:
I'm getting a SEGV when reading several rows with at least one LOB column. I get the first row but I get a SEGV in the fetch call when reading the second one.

LLDB stack trace:
```
(lldb) bt
* thread #6: tid = 0x941e0, 0x00000001054f3458 libclntsh.dylib.11.1`ttccfpg + 1368, stop reason = EXC_BAD_ACCESS (code=1, address=0x2c)
  * frame #0: 0x00000001054f3458 libclntsh.dylib.11.1`ttccfpg + 1368
    frame #1: 0x00000001054f01fe libclntsh.dylib.11.1`ttcfour + 8942
    frame #2: 0x00000001054cfd91 libclntsh.dylib.11.1`ttcdrv + 18433
    frame #3: 0x0000000104ca56b0 libclntsh.dylib.11.1`nioqwa + 128
    frame #4: 0x0000000104307051 libclntsh.dylib.11.1`upirtrc + 4321
    frame #5: 0x0000000104051c6f libclntsh.dylib.11.1`kpurcsc + 303
    frame #6: 0x00000001040367bd libclntsh.dylib.11.1`kpufch0 + 9005
    frame #7: 0x00000001040398c5 libclntsh.dylib.11.1`kpufch + 8757
    frame #8: 0x00000001042209f2 libclntsh.dylib.11.1`OCIStmtFetch2 + 98
    frame #9: 0x0000000101760895 oracledb.node`StmtImpl::fetch(this=0x0000000102905c00, numRows=<unavailable>) + 37 at dpiStmtImpl.cpp:426
    frame #10: 0x0000000101754adc oracledb.node`Connection::DoFetch(executeBaton=0x0000000102e053f0) + 28 at njsConnection.cpp:1682
    frame #11: 0x0000000101759f02 oracledb.node`ResultSet::Async_GetRows(req=<unavailable>) + 270 at njsResultSet.cpp:440
    frame #12: 0x0000000100656b2a node`worker + 90
    frame #13: 0x0000000100661a76 node`uv__thread_start + 25
    frame #14: 0x00007fff8cabb05a libsystem_pthread.dylib`_pthread_body + 131
    frame #15: 0x00007fff8cabafd7 libsystem_pthread.dylib`_pthread_start + 176
    frame #16: 0x00007fff8cab83ed libsystem_pthread.dylib`thread_start + 13
```

[...]

PR #211, opened 2015-09-15
fixed oracle/node-oracledb#210 - crash when fetching several rows wit…

bjouhier:
…h lobs

Problem comes from the fact that the descriptors are freed after the first row has been read. I fixed it by reallocating the descriptors before every row.

Not sure that this is the correct fix but at least it gets me going for now.

Note: my code executes a SELECT with `resultSet: true` and then reads every row in turn with `rs.getRow(callback)`.


cjbj:
This was merged to node-oracledb 1.2.  Thanks @bjouhier

PR #321, opened 2016-01-12
Support for streaming results for queries via new option (stream=true)

sagiegurari:
This pull requests provides ability to stream query results but also provides 2 more additional items:
1. ability to extend the connection object via javascript. most nodejs developers won't want to develop in C++ i believe and with this change now they can contribute to the oracledb library via js.
This could help a lot to this project and should provide a lot more contributions.
2. editorconfig file since the standards in this project is a bit different than most js projects.


cjbj:
@sagiegurari thanks for submitting.  Can you fill in the OCA as mentioned in [CONTRIBUTING.md](https://github.com/oracle/node-oracledb/blob/master/CONTRIBUTING.md)?

sagiegurari:
done

dmcghan:
@sagiegurari I agree that the driver could really benefit from more JavaScript code! :) To lay a solid foundation down in this area, I suggest an architecture where we create a JavaScript equivalent for each of the driver's base classes. End users of the driver would only work with the JavaScript classes which would wrap the underlying C classes. This would allow us to add our JavaScript code in a more organized fashion.

To show you what I'm talking about I've put together together this example:
https://github.com/dmcghan/node-oracledb/tree/javascript_classes_demo/lib

Keep in mind it's not fully baked, but it should get the idea across. In addition to demonstrating the architecture I'm suggesting, I've also added a connection request queuing feature for the Pool class. You can see how this works looking at pool.js and connection.js.

I'd be really interested to hear what you think about this architecture. Could you see yourself adding the streaming feature there? I've created a resultSet (resultset.js) class but right now it just calls through to the underlying class...

sagiegurari:
I couldn't agree more with the need and I see what you did in your fork.
You definitely put some much needed order in the code (talking about the lob and oracledb objects).

Basically it is similar approach in terms of extending oracledb like I did (take a look at another project of mine called simple-oracledb which I provide a lot more extensions) to provide additional logic in js.
However, while I extend the C++ object with JS, your approach is to create a wrapper object.
This which leads to the need of having every public c++ api also defined in the js.
My approach only requires the actual js modified api to be defined, and if a certain api there is no need to extend (lets say the connection.commit) than there is no need to define it.

The wrapper approach leads to more maintenance work, theoretically more bugs and in some cases might not work.
So i'm definitely for the goal you set, but the way it is set I think might be high maintenance.
Take a look at https://github.com/sagiegurari/simple-oracledb and you can see how much you can push via js to this lib without really wrapping it, and I have pushed a lot of features there.
It allows me to support any version of oracledb without doing any changes, even if oracledb add new functions to connection or pool and so on....

dmcghan:
That's a fair point regarding the need to manually add whatever needs to be exposed to the JS layer. Though for simple pass throughs, this is quite simple so I'm not so worried about bugs there. When you say, "and in some cases might not work" what did you mean?

Another issue I ran into with this approach is that properties like "pool.connectionsInUse" can't be mapped through very easily (at least I don't believe they can be as they are not like). Instead, getter methods would need to be introduced. 

On the other hand, this allows us to use JavaScript prototypes for all of the methods we'll be overriding. In the long term, as the code builds up, this could yield some memory benefits as there'll only be a single instance of the function on the prototype. Additionally, in one sweep, we could introduce a well laid out JS layer that folks could probably understand a bit easier and build out when needed.

Food for thought! :)

sagiegurari:
when I said might not work, is one scenario where in the native object it might call another public function which would result in bypassing the wrapper.
now... i'm not sure what will happen when you extend the object like I did as I never extended node modules via C++, but i'm hoping it will go to the js overridden function if called correctly in c++.
but i'm not sure.

dmcghan:
A native object calling an instance method seems unlikely - and as you say, it may or may not be an issue with either implementation.

Also, one of the enhancements I'd like to add to the driver is to make all async methods return promises over just undefined. This would mean that all methods (they're currently all async), including commit and release, would need to be overridden.

cjbj:
- I'm keen to see one or the other solution merged as long as it moves the driver forwards.
- The C layer can be changed as needed to support the JS layer
- The node-oracledb JS coding standard can easily be changed

dmcghan:
I rounded out the JS layer today in the demo branch:
https://github.com/dmcghan/node-oracledb/tree/javascript_classes_demo/lib 

I was able to figure out a solution for the the attribute values using Object.defineProperties. 

@sagiegurari Perhaps when your OCA is processed you could look at adding your changes there? Or I could do it for you?

dmcghan:
I rounded out the JS layer today in the demo branch:
https://github.com/dmcghan/node-oracledb/tree/javascript_classes_demo/lib 

I was able to figure out a solution for the the attribute values using Object.defineProperties. I think things are looking pretty good.

@sagiegurari When your OCA is processed I'd like to have your streaming changes merged to see how it all hangs together. What do you think?

sagiegurari:
I'm ok with doing changes whereever needed.
But why not use the power of js to change any object and simply extend instead of wrap? What is the advantage of wrapping?

dmcghan:
Normally we'd extend classes in JS using either prototypes (vanilla JS) or util.inherits (Node.js). The idea with these solutions is to extend out one constructor function (sub) via another constructor function (super). However, the driver currently only exposes the base class: oracledb.Oracledb. In all other cases, we only get instances of classes. To do things via prototypes or util.inherits, the C layer would need to expose all base classes. That's definitely worth having a discussion about. Definitely! :) But in the mean time, I think this is the next best thing. 

Our current approaches are actually quite similar. We are both sorting a reference to the C layer method before adding a JS layer in front. I'm just doing it wholesale which would happen anyway when adding the Promises.

sagiegurari:
not always. many times you augment an instance to provide more capabilities or change existing behavior. I believe that is what expressjs for example does with the nodejs request/response objects.
I'm just not crazy about wrapping every function and every attribute instead of just using the existing object and winning forward compatibility of any change.

cjbj:
@sagiegurari can you quantify 'not crazy'?  We'd obviously be keeping C & JS in sync going forward.

dmcghan:
I was wrong! Seems the base object DOES expose all the other classes (Oracledb, Connection, Pool, ResultSet, and ILob)!

I'll work on a new branch today that implements a better pattern using prototypes.

dmcghan:
Well, I tried, but I'm giving up on this approach (for now anyway). It was able to get this working with the base class. But the problem with this approach is that the other classes were not meant to be instantiated via "new" in the JavaScript layer. So, for example, when creating a pool you'd still have to call .createPool from the base class which would return a pool instance that wouldn't have any of the changes you made in the JS layer. :(

sagiegurari:
@cjbj "not crazy" :) means that javascript gives you the power to modify any object.
that's not something you see in other languages.
you have many libraries using this power to enhance other libs, for example I use it to change the nodejs spawn to first do something before the spawn happens.

Wrapping objects leads to more maintenance it has no advantage (only disadvantages) compared to modifying the existing c++ instances.
For example
1. all public apis MUST be in js object
2. all attributes MUST be in sync. If C++ changes it, js must also (I think @dmcghan is using define for that)
3. you might impact other libs (like mine) that modify the base objects (not sure....)
4. C++ should never call other C++ functions that are defined public as they will bypass the js layer
5. The biggest thing however, is that I can't think of any advantage of wrapping. It just doesn't give something more than modifying. The code from @dmcghan is very organized and properly written and gives a lot of order to the code for people to do changes, but why wrap? you can give order and help by modifying like I did and I don't mind you taking it forward and do it better than I did. But wrapping just doesn't give anything extra.

sagiegurari:
I think i can give a better solution, in my pull request, i specifically changed the execute function and kept the old one.
if you look what i did in simple-oracledb, there I did it in generic way for ALL functions, so no need to specifically state what function we want to extend.

In connection.js:
https://github.com/sagiegurari/simple-oracledb/blob/31a73553056a2ff206fa4e5f4c5b4c886c2f7e75/lib/connection.js#L838-L846

In pool.js:
https://github.com/sagiegurari/simple-oracledb/blob/31a73553056a2ff206fa4e5f4c5b4c886c2f7e75/lib/pool.js#L136-L144

If I would do the same in the connection.js of this pull request as I did in simple-oracledb, I think it should help give a solution that would satisfy all?

dmcghan:
I spent some time yesterday exploring a standard prototypal inheritance pattern that didn't work out. Let me spend some cycles today on an extend/mixin pattern. 

Frankly, I could go either way. What I really want to see is a release where each base class is extended in a consistent manner so that we have a good foundation for JS developers to build out.

sagiegurari:
:+1 i'll be happy to give any input once you have something to look at.

basically you can also modify the result set by modifying the connection.execute and wrapping the provided callback with your callback which would extend the c++ resultset with js level capabilities.
however i'm uncertain if that is needed.
I believe most developers would have good contributions mainly to the connection and pool classes and that only extending those 2 would be enough.
it is a classic 80 20 rule :)

dmcghan:
Okay, here's a first pass at an extends implementation: https://github.com/dmcghan/node-oracledb/tree/javascript_extends_demo/lib

Let me know if you what you think...

sagiegurari:
better, few comments that I think can improve:
* no need to write extender on every file. easier for people to have connection.js, pool.js and not extender. developers would look for connection.js not for connectionextender.
* no need for oracledb and oracledb extender, i mean, you are in the library doing the changes, so do it directly in oracledb.js, makes more sense and a lot more organized.
* just food for thought, i know it is intentional but keep in mind that empty functions that just do apply on the args have performance impact on the process due to 2 things.
one is the fact that you access arguments which takes time to create (only created if acccessed), second is that apply is slow. overall you won't see much impact since the IO to the DB would take much longer, but like I said, think about it.

overall i'm much for this change :+1: and if you need me to do after your pull request, another one to push the streaming I'll do it.

dmcghan:
Cool!

Regarding file names, I just want to be consistent. I started with a class=file approach. However, the extenders are not true classes (constructor functions invoked with the `new` keyword) which led to them being modules. Then the clash between the oracledb class and the extends module for it lead to those names...

If we take a standard Node.js module approach, grouping things by related functionality, and just expose the interface (constructor or extend functions) then we'd end up with something like this:
https://github.com/dmcghan/node-oracledb/tree/javascript_extends_demo/lib

I do like that better, what about you?

As for the empty functions, that was really just do demonstrate the approach we take when extending instances out. Keep in min that I do plan on building them all out so that they return Promises rather than just `undefined`. Those placeholders will help me when I go to do that (maybe this weekend?).

sagiegurari:
looks good.
as for promises, I suggest you take some library that implements it (there are plenty) and not es6 promise as you will want oracledb to support node 0.12 and 0.10

dmcghan:
Thanks! And thanks for your suggestions! :)

We would use es6 promises by default. If undefined, we could fall back on the es6 promise polyfill (a subset of rsvp.js):
https://github.com/jakearchibald/es6-promise

ecowden:
FYI: I've seen Node.js library authors leaning towards [pinkie-promise](https://github.com/floatdrop/pinkie-promise) as a promise polyfill. It manages the process of using es2105 promises when available and adds its own implementation only when not running in an es2015 environment. It's also very small, so it doesn't add a much extra weight to the library.

dmcghan:
lol, clever name! Thanks for the heads up, we'll have to look into both...

Bigous:
Well this is a very good job, congrats @sagiegurari.
Instead of writing it in pure JS, we could write the script part in Typescript. It gives us the advantage of write code with the newest features (class, annotation, etc...), generate type definitions for people that use the library to have code completion, hint and help while editing code and you end writing less code.
About the wrapping thing, I personally doesn't like it. I think that the code gets more difficult to maintain and to follow what is happening.
But I agree with you guys, people get afraid of C++ code often... Probably if the code were written in script more people would try to help to improve it.
I think that the best solution to be fast to maintain and have more people on it, should be a pure script driver (like oracle did with thinclient for java and oracle managed data access for .NET)...

dmcghan:
I added my thoughts on using TypeScript here: https://github.com/oracle/node-oracledb/issues/333#issuecomment-173055439

dmcghan:
@sagiegurari Chris recently commented about how the pool related logic in the connection's release method looked a little out of place. I agreed and thought the "correct" way to approach this was to make the connection an EventEmitter. That way, the connection could simply emit an event like "beforerelease" which the pool instance could use to do what it needs to do. 

The problem is that Node.js' EventEmitter requires one to use typical JavaScript classes which are extended via its EventEmitter class. Here's their example:
```javascript
const EventEmitter = require('events');
const util = require('util');

function MyEmitter() {
  EventEmitter.call(this);
}
util.inherits(MyEmitter, EventEmitter);

const myEmitter = new MyEmitter();
myEmitter.on('event', function() {
  console.log('an event occurred!');
});
myEmitter.emit('event');
```

With the new (mixin/extends) implementation of the JavaScript layer we don't have JavaScript classes which can be extended in this way. I'm not saying we should abandon this implementation, but you did ask for a reason why we might take the approach of a class based/wrapper implementation and I think we may have hit on a good one.

I'm still exploring work arounds for this issue, but I do worry about running into similar issues in the future.

What do you think?

sagiegurari:
I know how to extend event emitters, wrote an entire library on that (https://github.com/sagiegurari/event-emitter-enhancer)
I see no issue to support it with my implementation.
BUT, i'm not sure what pool issue you are talking about and if event emitter is what is needed.

dmcghan:
Let's say we want to make the connection class an EventEmitter (could be any of the classes really) and emit a "beforerelease" event when a connection is released. How would we do that?

sagiegurari:
I'll have to ask why, but to answer i'll say that you can inherit at any point in time and you do'nt have to do it in the constructor of your object.
the beauty of js.
but seriously, I'll have to ask why. I also did few pool extensions in my lib and I saw no need for an event emitter. Now... i'm not sure what you have in your fork, but i doubt if you need it.
the best way as i see it is to modify the release function to call the pool (i'm guessing of the need for an event).

dmcghan:
The motivator is just a separation of concerns. If the release method emits an event then it doesn't need to have a reference to the pool and call pool methods. The pool's getConnection method can be modified to add an event listener to the release event instead.

sagiegurari:
if the pool created that event, it should modify the release function.
The implementation of that release function is defined in the pool and it will override the default connection.release.
so connection class doesn't really know of the pool, but the pool does know how to modify the connection.
I think that is the proper way and it will be prevent leaks (you don't really want that someone will clear all listeners right?)

dmcghan:
Okay. I just used a simple workaround where the connection.release invokes a pool method. The latest version of the fork can be seen here: https://github.com/dmcghan/node-oracledb/tree/javascript_extends_demo/lib

sagiegurari:
* simple way to make the connection also an event emitter (you can do that in the Connection.modify function in my fork):
````js

    var EventEmitter = require('events').EventEmitter;

    Object.keys(EventEmitter.prototype).forEach(function (name) {
        if (typeof EventEmitter.prototype[name] === 'function') {
            connection[name] = EventEmitter.prototype[name];
        }
    })
    EventEmitter.call(connection);

    connection.on('test', function () {
        console.log('and.... it works!!!!');
    })

    connection.emit('test');
````

* I'm still not sure you want the connection to have the pool class.
I believe that you can override the getConnection in the pool and before calling the callback with the connection, you can override the connection.release, do your pool related logic there (again it is defined in the pool.js) and than in the overridden connection.release at the end, call the original connection.release.

````js
////pool.js getConnection changes, never tested but solution should work :)

pool.getConnection = function (callback) {
  //call base get connection
  var connection = ..... //im skipping this code to make example shorter.... and ya i know it should be a callback to get the connection....

  var orgRelease = connection.release;
  connection.release = function (cb) {
     //write pool release stuff here, no need for events.
    pool.dosomething();

    //call original release
    orgRelease.call(connection, cb);
  };

  callback(null, connection);
};

````

**Comment by dmcghan** [comment]
Yeah that would work. But we'd have to do all that work each time we get a new connection. With classes you'd just do it once on the constructor function and new instances just get the methods.

Ultimately, for this, it's not such a big deal. I went with something very close to your second suggestion. Hopefully we will not run into a situation where it does matter. I was just pointing out that there some drawbacks to either implementation. Just something to be aware of really...

**Comment by sagiegurari** [comment]
any updates on the OCA approval? I see another pull request also stuck on that.
I resent it on 24/jan.

**Comment by cjbj** [comment]
@sagiegurari I just pinged the OCA team.
Regarding this PR, we are evaluating Dan's changes for 1.7 (target this month). It would make sense to rebase on 1.7 when it is out?

**Comment by sagiegurari** [comment]
ok

**Comment by cjbj** [comment]
@sagiegurari the question is: Is @dmcghan heading in the right direction?! :)

**Comment by dmcghan** [comment]
Good thing I don't follow this thread. Wait...

**Comment by sagiegurari** [comment]
I guess the latest is his javascript_extends_demo branch.
it needs few refinements, small stuff like 
1. why do require with __dirname and not ./
2. why connection has reference to the pool
But in general I think its a good direction.

**Comment by sagiegurari** [comment]
oops sorry @dmcghan saw your reply after sending

**Comment by dmcghan** [comment]
lol! No worries... :)

We've already addressed "__dirname" vs. "./". It's now "./" throughout. 

The reference to the pool is added so that when the connection's release method is invoked it can call the pool._onConnectionRelease method (we discussed this above). It has to do with the connection request queuing.

**Comment by sagiegurari** [comment]
ya, i still think the pool needs to change the connection.release function before providing the connection. it is more correct flow.
that way the connection doesn't know, doesn't care.
but it is not critical i guess

**Comment by sanfords** [comment]
Hello, I was asked to comment on this branch.

_To the end of `.editorconfig`_: yes, editorconfig should be used for **every** project, not just this one.  :-)

_To the end of a streams API_: We do a lot of time consuming work on the data we select from the database, so streams are not the best solution for us.  This is mostly because of the throttling issues. The input would certainly overwhelm our ability to process the data. 

Personally I prefer the predictability/conformity of node-async and the current getRows implementation over promises and events.  Over the years it seems every library has idiosyncrasies around 'finally', 'end', 'close', 'finished', 'whatever', and determining what's called when is always confusing.  This makes it hard for us users to write maintainable code with them.

**Comment by dmcghan** [comment]
@sanfords As a general rule, I think we should try to adhere to Google's JavaScript Style Guide: https://google.github.io/styleguide/javascriptguide.xml

Of course we can put some of that into a config file, but not all.

What do you think?

As for the steams API, I really appreciate your thoughts on that - steams can be a pain! :) Of course, for some folks, steams might be more intuitive than recursive functions... I think the idea here is to allow the developer to choose which one they want to use. The steams API is just syntactic sugar on the standard ResultSet.

**Comment by jeffm13** [comment]
@dmcghan, I like the approach you took with your extensions.  I would, however, make one suggestion: Every async function in your extension should make the callback optional.  If the callback is undefined, return a promise. We wrap all oracledb interaction with promises.  If the extension did that for us, we could eliminate most of our own custom extension.   Taking this approach would mean that users could choose to take either approach: if you want promises, don't pass a callback. If you don't want promises, pass a callback. 

In terms of the streams API: Absolutely, a resultset stream would be useful.   We've implemented resultset processing with recursive promises.  It took some work, but it isn't rocket science.  But now we're about to go through an exercise where we will need to process hundreds of millions or even billions of rows.  Our currently implementation won't scale to that volume. Streaming would likely simplify it immensely.

**Comment by jeffm13** [comment]
Another option for style: 

[https://github.com/feross/standard](https://github.com/feross/standard)

Although it does take the fun out of arguing about style guidelines... Adding standard to the test suite, and enforcing successful tests on pull requests guarantees that they can't be violated.

**Comment by dmcghan** [comment]
@jeffm13 Optional Promise support is planned. We are exploring options that allow folks to use their own flavor of promises too since that can be such an interesting topic: https://github.com/nodejs/node/pull/5020#issuecomment-178112150

**Comment by dmcghan** [comment]
@jeffm13 No semicolons?!? What kind of archaic world do we live in? ;)

**Comment by sanfords** [comment]
> As a general rule, I think we should try to adhere to Google's JavaScript Style Guide

I like the 'BE CONSISTENT' part.  :-)  Beyond that, whatever makes the primary contributors happiest is fine.  For the moment, I'm more of a user.

We're in the same scaling boat.

**Comment by sanfords** [comment]
> Every async function in your extension should make the callback optional. If the callback is undefined, return a promise.

+1, for us callback users.

**Comment by sagiegurari** [comment]
The editorconfig is not for coding standards, it was because the spaces amount in this project is not consistent with other js libraries and it is hard to have to fight an IDE which wants to reformat your code all of the time. So... editorconfig fixes that :)

if you really want to define standards, you will need to setup jscs/eslint/... config files and have a proper build for this library which enforces that.
That's what I do in all my projects.

As for promises, I agree that it should only come if callback is not provided, so users won't be forced to use promises (which kind of make it hard to use the async library).
But that is not part of the scope of this PR.

For those that don't need streams, keep in mind that it is optional and it is not the default behavior.
You must provide the option stream=true for the execute to return a stream, so it should not impact users that don't need/want it.

Anyhow, once this PR goes through and this library can be extended via js, than we can start contributing more to node_oracledb.
But I admit that the releases are too far apart and it takes to long to push things in.
I think every little change can be released with a new patch version (x.y.[patch]) and once every few weeks, a new minor version (x.[minor].0).

**Comment by cjbj** [comment]
@sagiegurari check out the changes in 1.7.  Looking forward to seeing this rebased.

**Comment by sagiegurari** [comment]
ya, hopefully I'll be able to do that next week.

**Comment by sagiegurari** [comment]
I merged the changes of 1.7.1, hope it is ok now.
Please run also some tests on your end to see if it is ok.

**Comment by dmcghan** [comment]
Thanks Sagie!

Just a couple questions to get the ball rolling...

- Since execute is an async function, it's important that it not return a value. This will allow us to add the optional promise support where all async functions will start to return promises. Could we just return the stream in the callback's result?
- Since we went with a module based file convention, would the ResultSetReadStream code be better off in the resultset.js file?

**Comment by sagiegurari** [comment]
1. return value - actually that's a feedback from the original thread which I happen to agree with. meaning you get a read stream in the response and right away start listening to data/error/... events.
it is replacing promise in case you want streaming.
think about node.js net connect function which is async but returns the socket right away so you can right away start listening to data/error events but the callback is optional.
so it is more node.js way and does not hurt the promise changes in the future (meaning if stream, the return value is stream instead of promise).

2. modules - If a class can be self contained than it is preferred to help minimize file size and complexity.
it is recommended to have 1 class per file where possible.
if you look at the code, it actually doesn't change anything in the core oracle resultset class, it is just a stream on top of resultsets, just like a connection is on top of resultset, and pool on top of connection.
So having small self contained modules is better, easier to test, easier to read (read is important).
it is not always possible/easy to split to small files, i admit.

**Comment by dmcghan** [comment]
I think we need to be careful when deviating from a standard (especially with such a common, generic method like execute). Currently the async functions done return anything. This would change that to return a type based on a parameter. Could this happen again, say for AQ or CQN? What about combinations of parameters that would compete with each other?

We have to consider this independent of but in tandem with the promised based API. There are various ways the optional promise support could be added...

1. If no callback function is supplied then we return a promise. 
2. We add new methods that follow a pattern, like executeAsync, that return promises.
3. We could expose a promise/promisifed method on the current method that works with promises. That seems to be the direction they are leaning here: https://github.com/nodejs/node/pull/5020

Which of these are you most in favor of and or against? How will this impact return types in general?

**Comment by sagiegurari** [comment]
I don't think promise will conflict with the return of stream.
I think that is taking advantage of the power in js (compared to langs like java).

no callback -> return promise - this one is ok, unless you asked for stream and than instead of promise, you get a stream. best tool per request. if in future we add more patterns, i think that should be ok as well, for example see simple-oracledb query has 2 options that conflict but I have priority (stream and split).

The third option is really interesting. I need to think about that one more, but that could really be nice to have '.promise' which actually outputs the original api after being transformed to promises with something like bluebird.
Could make the api more understandable than the 'no callback' approach.
I have to admit, I like this approach the most.

**Comment by jeffm13** [comment]
> no callback -> return promise unless you ask for stream, then it returns a stream 

This feels a bit inconsistent.  Maybe a fourth option would be to add a `resultStream` property to the `results` object (much like `outBinds`--the property only has a value if it needs to).  We're doing something similar in our wrapper to support smaller `resultSet`s. 

When wrapping with promises, the `execute` method could return a promise that resolves to either a standard results object or to a stream if requested.

**Comment by sagiegurari** [comment]
@jeffm13 you are suggesting instead of
````js
var stream = connection.execute(sql, bindings, options);
stream.on('data', func);
stream.on('error', func);
//.... and so on....
````
To do this?
````js
connection.execute(sql, bindings, options, function (error, results) {
  if (error) {
    //also handle error here
  } else {
    var stream = results.resultset.stream;
    stream.on('data', func);
    stream.on('error', func); //another place to handle errors
    //.... and so on....});
  }
});
````

To me, this looks long and complex compared to just get the stream right away and start using it.
See nodejs example for similar pattern:
https://nodejs.org/api/net.html#net_net_connect_port_host_connectlistener

**Comment by cjbj** [comment]
A couple of quick comments: (i) somehow close() needs to be called and/or callable. (ii) I am seeing a silent failure if the options has `{stream: true, outFormat: oracledb.OBJECT}`.

**Comment by sagiegurari** [comment]
@cjbj you are correct about the first item. I didn't migrate the code as is from simple-oracle which has the result set reader that makes sure to close the result set, so that's a bug I'll fix today.
As for item 2 I'll try to reproduce and fix if needed.
Thanks

**Comment by cjbj** [comment]
@sagiegurari For performance, what effort would it be to use getRows instead of getRow?  Maybe overload the use of maxRows as the getRows parameter?  I'd be interested in seeing some benchmark results using getRow vs getRows

**Comment by sagiegurari** [comment]
getRows will be better naturally. I'll investigate supporting that, shouldn't be an issue I think.

**Comment by jeffm13** [comment]
@sagiegurari Have you thought about instead creating an executeStream function instead that serves as a stream factory for a statement returning a resultset?  So it would look like: 
```javascript
var stream = connection.executeStream(sql, bindings, options);
stream.on('data', func);
stream.on('error', func);
```
That way the package isn't mixing semantics: an async function takes an optional callback and returns an optional promise. The user chooses how to deal with async behavior: with or without promises.  A stream factory creates a stream and the user deals with it using standard stream patterns.

sagiegurari:
@dmcghan and @jeffm13 since both of you raised concerns about promise and overloading the execute, I see no reason not to change it to another function.
I'll probably call it 'stream' and not executeStream.
Anyone thinks differently?

To me it is not critical, i'm not a fan of promises as I usually use the async lib instead (to me it is more readable), but since it raises concerns it is not a big deal to do that refactoring.

sagiegurari:
1. @cjbj I fixed the object binding issue and added a test case for that. thanks
2. @dmcghan and @jeffm13 I moved the new functionality out of the execute and into a new 'stream' function.

I have one issue with the resultset close which I need an advice about.
In simple-oracledb, when I stream, I also convert all CLOBs and BLOBs into string/buffer so the user on the 'data' events, gets a row object with data and not streams.
The connection can be closed and it wouldn't matter, as the user now has the actual data in a JS object.
However in this PR, I provide the row object as is in the 'data' event.
The result is, that if I read all rows and provided those rows objects via 'data' events, and than I close the resultset, I believe users that will only than read the LOBs will fail as the resultset just closed.
So, there is no way for me to say when I can close the resultset.
What I'm thinking about, is to move a bit out of the basic stream api, and add a 'stream.close()' function which will than call the resultset.close() and than the users will have to call that close after they finished with the stream.
What do you think? any better solution?

Second issue is the getRows by @cjbj which I see issues since I'm getting errors from oracle on unable to run concurrent operations on a resultset.
So i'm not sure I'll go for it now, but we can always add that support at later time.
Anyway, I'll play with that a bit more this week to see if I get any more insights.

cjbj:
I've been thinking about this a bit; and will discuss it with @dmcghan and the team tomorrow.

- If you're going to have a separate method, what about calling it queryStream()?

- Wouldn't there be issues if a queryStream.close() method was called prior to the end of the stream?
 (Isn't it a Streams limitation that Readable streams have to be completely fetched?) So a .close() might cause more confusion when misused, and generally be yet-another-method to remember the semantics of. Personally, based on the 80-20 rule, I'd be OK with LOBs being instantiated into a string/buffer and the resultset.close() being executed internally on normal success and failure code paths.  If users don't want LOBs instantiated, they could implement their own streaming. However I could be out voted. 

- some test ideas: 0 rows, 1 rows, 2 rows, non queries, LOBS

- pity about getRows. Keep trying!

sagiegurari:
1. queryStream - done
2. close - see fs readstream.close/destroy functions, seems it is an acceptable option
https://github.com/nodejs/node/blob/4ed038808f7bb78cc00fce6162b3f71b45abbd66/lib/fs.js#L1884-L1889
3. tests - will do
4. getRows - Haven't given up yet. I'll write more tests to check this out.
If I get a specific scenario, maybe you can help out or something.

dmcghan:
@sagiegurari I read through that Promise related thread for Node.js core. It seems they moved away from returning a promise if a callback was not supplied because of some minor issues with their existing API (they already had some functions with optional callbacks). Additionally there are some changes in the way that programmer vs. runtime errors are raised or passed asynchronously. It seems the debate lead to them abandoning the PR altogether for now.

At any rate, I don't believe we suffer from these same issues and I think reusing the existing methods for optional promise support will result in the simplest API. I have a basic promisify function working. I'm going to test it some tomorrow, along with your branch to see how they fit together. I think I prefer the original execute method with a flag over a new stream method just for this one variation on a return type. Sorry, I didn't mean for you to change anything - I was just throwing out ideas...

Regarding resultsets and close, have you tested streaming a lob after the resultset is closed? If not, I can do that tomorrow. I'll also take a look at the getRow vs getRows codel...

sagiegurari:
Thanks, I am going to test it, just a bit busy lately, but logically it should not work.

As for the execute with flag or new function, I don't have a strong opinion about that, I preferred 1 smart execute function (like in simple-oracledb connection.query) but since it is not my project, I'll let you, chris and others decide.

sagiegurari:
ok, got some good news.
Seems that I am able to read from a LOB object even if resultset is closed (don't ask me why).
So... I'm closing the resultset after we are done reading automatically and no need to provide a stream.close function.

Also I added more tests per @cjbj request, including 1 test that checks CLOB reading AFTER the resultset was closed.

Please review.
If you want to move it back to the execute, I can do it very easily.

Next I'll have to look at the getRows one more time, but that is internal implementation and can be changed at any time later on.

sagiegurari:
ok, now I added the support for getRows instead of getRow.
The rows fetched in each iteration is based on maxRows options as follows
````js
var maxRows = oracledb.maxRows || options.maxRows || 100;
````

As I see it, all that is left is to decide if it is part of execute function or a new function.
So i'm waiting for a decision on that.

**Comment by dmcghan** [comment]
@sagiegurari I'm happy to hear that LOBs worked and that you were able to get getRows working too!

I forked your fork and played with it some today. I even moved the streams logic back to the execute method. However, I think the way you have it now (using a separate method to get the stream) is better overall.

I submitted a pull request to your fork with some minor stuff.

One thing that's bothering me is that users of the stream API can't get to the metaData (which may be extended in the future). At first I thought it might be as simple as adding a custom "metadata" event that would fire before the data itself. But now I'm not so sure... Do you have any ideas on how to support metadata? Maybe an options property that, if true, sends the metadata as the first row in the stream? What do you think?

**Comment by cjbj** [comment]
@sagiegurari Great progress.  I can see the perf improvements from
larger resultSet batch sizes (but see the `maxRows` buglet mentioned below).

Re LOBS and `close()`, the devs pointed out that in OCI you can access
LOB locators when the statement handle is closed as long as the
connection is open.  It's just a matter of making sure the
node-oracledb C layer doesn't do some cleanup.  Thanks for adding a
test to ensure this.

Comments / Open Question:

  - Reuse `execute()` or a new method e.g. `queryStream()`?  Tending
    towards new method.

  - Overload maxRows or have a new attribute?

  - Query metadata: is it needed?  Can it be accessed via an event if
    not using `oracledb.execute()`?  I'm not so keen on Dan's
    'optionally make it first row' idea, but I admit it is a practical
    solution.

  - Coding standards (what's wrong with Emacs indentation!?)  Comment
    blocks, attributions etc.  Need to fit with current standard.
    We can sort this out during the merge.
    FWIW, I'm going to belatedly retrofit CHANGELOG with contribution
    attributions; these were previously in the release notes of each
    release.  

  - Future Promise use.  Dan will likely post separately how it might
    look.  We need to keep it in mind when thinking about the query
    streams API.

  - A user app's oracledb.maxRows value is ignored due to the
   ` require('./oracledb'`) in connection.js.

Doc in api.md (I can tidy up when merging):

  - mention must fetch until the end else the RS close won't occur and
    the app leak cursors.

  - remove `<br>` tags from api.md.

  - mention `maxRows`

  - Doc obviously needs updating when final names etc have been decided.

Tests:
  - Potential tests:
    - `maxRows` values (perf test)
    - pause/unpause and other streams functionality

  - Stress testing.  We want to be able to do some solid load tests.
    Time is tight this month.  We'll do our best.

**Comment by sagiegurari** [comment]
Thanks @cjbj and @dmcghan 

*  reuse execute - because many people use bluebird to convert the api to promises, I also tend to agree with people here that it should be a new function. So lets keep it as it is.
*  overload maxrows - maxrows might be confusing, I agree. we can create a new property bulkSize or something and if not available only than default to maxrows.
*  metadata - dan also suggested a new event for the metadata. I agree with that solution, that way people using the api, don't have to ignore the first record. So i'll create a new event that the stream will emit ('metadata')
*  coding standards - should be in a different PR. the editorconfig is more for IDE setup since the spaces are different than most projects. but standards should be defined via jscs/eslint/...
I'm all for it, but different PR.
*  promises - should be different PR. if the streaming stays in a new function, it should not cause any conflicts. Also the new function would not need to be promisified as it returns a stream right away.
* require('./oracledb') - dan fixed that with his PR which I merged. So we are ok with that.
* Docs - I'll modify the docs as you requested, however the docs right now are numbered everywhere and it is very hard to maintain. I prefer creating bullets via markup and linking to sections via name and not number. It makes life much easier.
The numbers you have are everywhere, meaning in docs, in some txt file in the tests, the tests themselves are also numbered. It really makes it hard for new functionality to be added.
Just something to think about for the future.
* perf/stress testing - I'll add some tests for that.

**Comment by cjbj** [comment]
For the batch size, perhaps `streamNumRows` (similar to `getRows(numRows,..)`)??? I'm open to ideas that people won't mistake for an upper bound of the number of rows queried. I'd keep the behavior decoupled from maxRows to avoid 'magic': either the new property is set, or a default of 100 is used.

Someday api.md should become api.html. For the moment we get the joy of shifting numbers.  Don't sweat the details.  I can tidy up as needed.

**Comment by dmcghan** [comment]
@sagiegurari @jeffm13 @ecowden I just pushed an branch that demonstrates the current direction we're looking at going to add promise support. https://github.com/dmcghan/node-oracledb/tree/optional_promise_support

There's a new promisify function in a new file named util.js. You'll see it's used to wrap all the async functions. The promise implementation can be swapped out by setting a new property on the base class: `oracledb.Promise`

@jeffm13 I think you mentioned you have a wrapper that does this. Is there anything that's missing in this implementation when compared to yours? Does it work the same?

Could any/all of you have a look, maybe kick the tires some? Any feedback (good or bad) would be appreciated. Thanks in advance and have a nice weekend!

**Comment by sagiegurari** [comment]
I'll check it out, but need to make sure the library supports both promise and non promise in parallel so if i'm using a library that expects a non promise version while I use a promise version than they both work at same time.
That is why I liked the idea of oracledb.promise that you linked to with that nodejs PR.

**Comment by dmcghan** [comment]
Thanks Sagie! 

Yes, couldn't agree more... The current promise support is based on the fact that the only functions passed as parameters to any/all class methods are callbacks - and they're always passed last.

So now when you invoke a method like execute, if the last parameter is a function, then things will continue to work as they always have. However, if it's not a function, then you'll get a promise returned.

I ran all of the existing tests (which use callbacks) and they all pass except 3 which are all parameter assertions (expected to fail given the changes).

**Comment by jeffm13** [comment]
@dmcghan, at first glance, this looks excellent. I believe everything that we need promisified, is.  It looks like it will allow us to eliminate about 90% of our wrapper; possibly we'll even eliminate it entirely.  I'll play around this weekend and see how it works out. 

The only things I see missing at this point is documentation and tests, maybe some examples, Oh, and then there are those parameter assertions :)

**Comment by dmcghan** [comment]
Thanks Jeff, that's great to hear! 

I did add a new file for tests named promises.js (though I'm sure it needs some work). Once I get some reassurance that things work "as designed" I can look into documentation and examples. 

Looking forward to hearing about the results of your testing!

**Comment by ecowden** [comment]
Initial smoke testing: things work as expected. Kudos and thanks -- this was a big one for me! :+1: 

Some truly minor nitpicking: 

* I'm wary of micro-optimizations without measuring, but is there a need to transform `arguments` into a proper Array [here](https://github.com/dmcghan/node-oracledb/blob/f0d69b4e6609ee376bf79211af5e5e5d087c6fa0/lib/util.js#L35)? The `args` variable is only used to get the length and the last element, and the code has already called `arguments.length` at that point. Might as well save the overhead of allocating an extra Array.

**Comment by dmcghan** [comment]
@ecowden Thanks for the feedback! :+1:

**Comment by dmcghan** [comment]
I just did a blog post that sums up the current status of this PR (ResultSet Streaming) and the promise support (that doesn't have it's own home yet - sorry Sagie!):
https://jsao.io/2016/03/extending-the-javascript-layer/

**Comment by sagiegurari** [comment]
The promise code changes look good to me.
Didn't run the tests but it seems to look simple (which is good :+1:)

**Comment by dmcghan** [comment]
Thanks for taking a look Sagie!

**Comment by sagiegurari** [comment]
ok so I 
* added the metadata event
* changed the getRows option used to streamNumRows
* Updated docs

Missing
* performance/stress tests

Hope to setup some tests today/this week.
if there is anything still missing please tell me.

**Comment by cjbj** [comment]
Thanks @sagiegurari.  I know @dmcghan wants to pound on it a bit.  We will also work on tests too.
When we merge, the comment format will change.

**Comment by cjbj** [comment]
To set expectations, Promises will be post 1.8.s

**Comment by sagiegurari** [comment]
Added a stress test of 500 queries using single connection with random streamNumRows value.
I guess I'm done with the PR coding, so if anything is needed just tell me.

**Comment by cjbj** [comment]
@sagiegurari that's a good point to stop so we can work with a stable PR. Thanks!

**Comment by cjbj** [comment]
@sagiegurari do you feel it's OK to emit a `close` event but not an `end`?

**Comment by sagiegurari** [comment]
I only emit the metadata and error events.
the nodejs stream code is the one that emits the other events based on if I pass data or not.

if you look at the following doc, you will see that 'close' is an optional event.
https://nodejs.org/api/stream.html#stream_event_close

I can add code to emit that as well, but i'm pretty sure that error and end are doing the job just fine.

**Comment by cjbj** [comment]
Aren't you emitting `close` here: https://github.com/oracle/node-oracledb/pull/321/files#diff-259dec4414a400a6e895f16ff1d0ca3bR53

My tests show that `end` is also emitted - and you use this in tests too.

Can you check the following, which is crashing for me.
```
    var stream = connection.queryStream(
      'SELECT first_name FROM employees ORDER BY employee_id',
      [], { streamNumRows: 100 });

    stream.on('error', function (error) {
      // console.log("stream 'error' event");
      console.error(error);
      return;
    });

    stream.on('metadata', function (metadata) {
      // console.log("stream 'metadata' event");
      console.log(metadata);
    });

    stream.on('data', function (data) {
      // console.log("stream 'data' event");
      console.log(data);
    });

    stream.on('close', function () {
      console.log("stream 'close' event");
    });
    
    stream.on('end', function () {
       console.log("stream 'end' event");
    });
```

sagiegurari:
* my code doesn't emit the 'close', it doesn't emit 'data' or 'end', it is the nodejs stream code that does that. It is triggered when I do push(data) or push(null) in my stream class.
* I run tests on multiple node.js versions. I can tell you that node 0.10 behaves very very differently than 0.12 and up.

cjbj:
I recall the issue: because there is no explicit use of the connection it gets closed before all data is fetched. If I add a connection.release() to the end event, all is well.

sagiegurari:
can you explain? is it a test error?

cjbj:
@sagiegurari no problem with tests/stream.js.  It was a problem with my standalone test closing the connection too early.

In other news to keep you in the loop, we're getting this PR stable in our internal repo for stress testing etc. Dan's fixed an issue with pooling due to the pool connection.extend not passing oracledb. I'm still trying to find a better name than streamNumRows.  We will likely move some functionality from connection.js to resultset-read-stream.js.  I've done a some work on doc.  Dan was wondering whether highWaterMark should be settable - this could be a post 1.8 topic to discuss.

sagiegurari:
Why use an internal repo and a branch per release that is open and visible to everyone? You will get better feedback early before you release it officially.

dmcghan:
@sagiegurari I think the discussion on a close/destroy method got lost. I think really important as without it we can't stop the stream, we can only pause it which keeps the resultset open (which blocks connection operations, like release). 

As there is no close method, I'm confused by this `if` block here: https://github.com/oracle/node-oracledb/pull/321/files#diff-259dec4414a400a6e895f16ff1d0ca3bR51 

How would the closed property ever be set?

sagiegurari:
That code is from when I did develop the close but removed it. Should remove that as well.
Wouldn't connection.break solve this?

dmcghan:
I just ran a test and break does work. However, I think it should be possible to stop the stream without breaking the connection. A close method seems the most appropriate way to do this, no?

sagiegurari:
I'm ok with that, I'll add it next week

cjbj:
@sagiegurari we've frozen 1.8 ready for release testing, so hold off making changes for the moment.

sagiegurari:
sure

cjbj:
@sagiegurari thanks for your perseverance.  We merged this to 1.8.  @dmcghan moved more of the code into the new js file. At the last minute we decided to revert to using oracledb.maxRows for the getRows() size.  We want to revisit this when https://github.com/oracle/node-oracledb/issues/361 is looked at.
Let's use new issues/PRs for the future work such as a way to close streams early.  Also I see getRows() is called one extra time; maybe this can be optimized.

jgoux:
Hello,
I tried to use this new feature, but I can't `pipe` the stream. Is `pipe` supported ?

Issue #169, opened 2015-08-17
will providing typescript type definition file in the plan?


haiwen-shaw:
I think using javascript is not good for a complex project, and the typescript seems pretty awesome for handling it. but it need .d.ts file to call node-oracledb. will oracle provide typescript type definition file (.d.ts) ? thanks!

Bigous:
I have just added a pull request to DefinitelyTyped/DefinitelyTyped#7697 to add oracle.d.ts
[]s

cjbj:
@Bigous Nice one!

Bigous:
Pull was merged... I think we could close this one.

cjbj:
Now, where should this be documented?!

Bigous:
DefinetlyTyped is a facilitator only. It holds files .d.ts that original projects do not release or maintain. I think that if it's gonna be there, we could only mention it on README.md. Something like:
-    For those who are using typescript, there is an updated oracledb.d.ts at DefinitelyTyped (tsd).

If we choose to bring it to oracledb itself on next release, we should add an index.d.ts at the root of package with the content of oracledb.d.ts. Then I think we should mention it on o README.md and improve it with the content of API.md (complete the code documentation).

Bigous:
Found out (angular2 project is an example) that we can have a typings directory inside our repo and install tsd dependencies on it, like node.d.ts, and have it all packaged here...
@cjbj If you'd like, I could move the definitions to this repo ... it occurred with angular2 too... definitions until version alfa-something is on tsd and after that is on main angular2 repo...
So, we could have the definitions for version 1.5 at tsd, and after that, here.
What do you think?

cjbj:
@Bigous I don't see a problem with adding it to node-oracledb.  Submit a PR?

Overall, we need to find a way to machine-generate API doc. I know @dmcghan was enthused today by http://documentation.js.org/

dmcghan:
Yes, @jed let me know about it today! I like that it covers the C/C++ and JavaScript (ES6).

@sagiegurari what do you think? You use JSDoc to Markdown, right? [documentationjs](http://documentation.js.org) seems to use the same syntax...

@cjbj mentioned having both reference docs and user guides. This could help automate the reference docs.

wallride:
definitions for oracledb module on DefinetelyTyped are for version 1.5, though the actual module version is 1.9.3 (supporting promises etc).
Will the actual typescript definitions be mantained by oracle?

cjbj:
[...]

Issue #333, opened 2016-01-19
Rewrite in typescript


Bigous:
Hi @cjbj , I'm working with angular2 for while now and having the typescript definitions is awesome - code completion, hints, etc... I wrote an oracle.d.ts and made a pull to TSD people DefinitelyTyped/DefinitelyTyped#7697 but I was thinking that it would be great to rewrite the javascript part of the driver in typescript and expose the definitions and so on, so we don't need tsd to have the oracledb definitions... and it would be generated automatically by the typescript transpiler...

We have an issue #169 asking for that too...

I could halp with it, what do you think?
[]s

cjbj:
@Bigous Sounds cool, but is there enough JS to merit another layer of abstraction?  If it makes development and debugging and maintenance easier, and attracts contributors then we could do it.  (BTW, any comments on https://github.com/oracle/node-oracledb/pull/321#issuecomment-172214766 ?) 

@dmcghan any comments?

Bigous:
@cjbj Typescript is Javascript with types... When I submitted the issue I was thinking about having just the 2 javascript files written in typescript, so the driver would get easier to use (because the typing and help would be at the script part which would be bundled with the driver... I was not thinking about having it entirely in Typescript... But I liked the idea.

dmcghan:
Generally speaking, I like TypeScript. I like what it does and what it has to offer. For medium to large projects I can understand why a development group may chose to use it. However, the driver is mostly developed in C. Excluding the tests and examples there's only about 100 lines of JavaScript code so I'm not really sure what, if anything, TypeScript would have to offer at this point.

Also, one of the main motivators of expanding the JavaScript layer is to allow the JavaScript community more opportunity to contribute to the project. There are a lot of people that know JavaScript but TypeScript has not become mainstream, and like CoffeeScript, it may never become the norm. 

My vote would be to stick with plain old JavaScript for now. If TypeScript ever becomes the norm or if the project ever gets to an unruly size, then I'd revisit this issue. 

In the mean time I do think we should provide and maintain a .d.ts file for the library.

cjbj:
@Bigous a thin driver (either in pure JS or typescript) is a dream but not a plan.  Node-oracledb will be based on Oracle Client for the foreseeable future.

Bigous:
Agreed with both... about .d.ts, I posted one in tsd repository... if you thing we should move it to be here, I can make a pull request here.
[...]

PR #346, opened 2016-02-01
Added index.d.ts to expose types for typescript.

Bigous:
Resolving issue #169.


Bigous:
Hi @cjbj, I think we could cancel this pull request until we decide if we are going to keep DefinelyTyped or use typescript as a helper at JavaScript layer, so we can have this automagicaly generated (and so the api documentation).
What do you think?
[]'s

Bigous:
Just to attach things - we discussed it, ending in lets wait at #333.
From that point to now (7 months), Typescript evolved a lot. I really thin that it would be very nice to have the Javascript part rewritten in TypeScript generating the Javascript and adding definitions and documentation on it.

cjbj:
@Bigous thanks for the updated pull at https://github.com/DefinitelyTyped/DefinitelyTyped/pull/10323   I think that is the right home for the definition file at the moment.  You may recall I had some reservations about merging to node-oracledb and there are a bunch of other things (temporary lobs, longs etc) that need attention.  If you close this, we can alweays reopen it later.

Bigous:
Sure, lets keep it separated for now and reopen later or create another one !!!

PR #365, opened 2016-02-25
Update api.md

sanfords:
getRows link broken in "fetched using..." sentence.


cjbj:
@sanfords thanks for taking the time to create the PR.  For us to accept a PR you would need to follow https://github.com/oracle/node-oracledb/blob/master/CONTRIBUTING.md
Since that's a bit of effort for this one char change, I'm going to treat your PR as a bug report and will get it fixed directly (and the other two broken links I can see).

PR #386, opened 2016-03-22
extended metaData

leolmi:
# extend metaData  items

`metaData` array now exposes items with more properties:

- `name` : field name
- `type` : field data type (JS datatype)
- `originalType` : field data type (DB datatype)
- `size` : DB field size
- `precision` : DB field precision
- `scale` : DB field scale
- `isNullable` : field can be null

```javascript
connection.execute(
  "... SQL statement ...",
  [],
  function(err, result)
  {
    var columns = _.map(result.metaData, function(c)
    {
      return {
        name: c.name || '',
        type: c.type || '',
        originalType: c.originalType || 'undefined',
        size: c.size || 0,
        precision: c.precision || 0,
        scale: c.scale || 0,
        isNullable: c.isNullable
      };
    });
  });
```
Signed-off-by: Leonardo Olmi [leo.olmi@gmail.com](mailto:leo.olmi@gmail.com)

### DIFF
--- a/lib/oracledb.js
+++ b/lib/oracledb.js
@@ -87,6 +87,122 @@ function extend(oracledb) {
         value: 0,
         enumerable: true
       },
+      DBTYPE_VARCHAR: {
+        value: 1001,
+        enumerable: true
+      },
+      DBTYPE_NUMBER: {
+        value: 1002,
+        enumerable: true
+      },
+      DBTYPE_INTEGER: {
+        value: 1003,
+        enumerable: true
+      },
+      DBTYPE_DOUBLE: {
+        value: 1004,
+        enumerable: true
+      },
+      DBTYPE_STRING: {
+        value: 1005,
+        enumerable: true
+      },
+      DBTYPE_LONG: {
+        value: 1008,
+        enumerable: true
+      },
+      DBTYPE_DATE: {
+        value: 1012,
+        enumerable: true
+      },
+      DBTYPE_RAW: {
+        value: 1023,
+        enumerable: true
+      },
+      DBTYPE_LONGRAW: {
+        value: 1024,
+        enumerable: true
+      },
+      DBTYPE_UNSIGNEDINTEGER: {
+        value: 1068,
+        enumerable: true
+      },
+      DBTYPE_ROWID: {
+        value: 1104,
+        enumerable: true
+      },
+      DBTYPE_FIXEDCHAR: {
+        value: 1096,
+        enumerable: true
+      },
+      DBTYPE_BINARYFLOAT: {
+        value: 1100,
+        enumerable: true
+      },
+      DBTYPE_BINARYDOUBLE: {
+        value: 1101,
+        enumerable: true
+      },
+      DBTYPE_UDT: {
+        value: 1108,
+        enumerable: true
+      },
+      DBTYPE_REF: {
+        value: 1111,
+        enumerable: true
+      },
+      DBTYPE_CLOB: {
+        value: 1112,
+        enumerable: true
+      },
+      DBTYPE_BLOB: {
+        value: 1113,
+        enumerable: true
+      },
+      DBTYPE_BFILE: {
+        value: 1114,
+        enumerable: true
+      },
+      DBTYPE_RSET: {
+        value: 1116,
+        enumerable: true
+      },
+      DBTYPE_YEARMONTH: {
+        value: 1182,
+        enumerable: true
+      },
+      DBTYPE_DAYSECOND: {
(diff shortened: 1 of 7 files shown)


cjbj:
Thank @leolmi.  Did you submit your [OCA](https://github.com/oracle/node-oracledb/blob/master/CONTRIBUTING.md) paperwork?  Also I should get more formal and request the 'Signed-off-by:' mentioned in that doc.

leolmi:
Hi @cjbj, yes, I have already sent the document to oracle-ca_us@oracle.com.

cjbj:
Thank you.  They batch up requests so it may take a short while for them to process it.

leolmi:
thanks to you, it was a pleasure!

cjbj:
@leolmi can you add the signoff?

leolmi:
what happened? I have to send the OCA again?

cjbj:
@leolmi the OCA may take time if they didn't get back to you.
I was asking about the commit message/description needing a `Signed-off-by: Your Name <you@example.org>` - see [CONTRIBUTING](https://github.com/oracle/node-oracledb/blob/master/CONTRIBUTING.md)

cjbj:
@leolmi we had some discussions. How about:

1. Adding a new execute() options attribute to turn off metadata gathering.  This should improve performance for people who don't want metadata.  The default would be to always have metadata; this keeps compatibility.

2. We feel that creating new numeric constants in lib/oracledb.js would be better than using strings for the DB types.

leolmi:
@cjbj 
>Adding a new execute() options attribute to turn off metadata gathering.

I think it's the right thing.

>creating new numeric constants in lib/oracledb.js would be better than using strings for the DB types.

It's probably fair, in addition'm not sure that the lists that I have used are exhaustive.

cjbj:
@leolmi now the OCA is processed (yay!), will you be able to make the changes I mentioned?

leolmi:
@cjbj of course! For a few days I'll be busy in other activities, but as soon as I can I will make the changes mentioned!

leolmi:
@cjbj second commit:

- `originalType` is now a numeric property of metadata item which refer to the values of the `oracle.DBTYPE_XXX` enum;

- to disable metadata on result you can use `noMetadata` property of execute options:
```javascript
connection.execute(
  "... SQL statement ...",
  [],
  { outFormat: oracle.OBJECT, noMetadata:true },
  function(err, result)
  {
    // ...
  });
```

leolmi:
@cjbj I haven't done extensive testing with new implementations!

cjbj:
@leolmi I'll look at the changes in detail later.  The first thing that catches my eye is the `noMetadata`.  Having negation in the name can be hard for people to understand.  Why not have `metaData: true` as the default? Therefore developers would explicitly use `metaData: false` to disable it?

leolmi:
@cjbj I agree

- to disable metadata on result you can use `metaData` property of execute options:
```javascript
connection.execute(
  "... SQL statement ...",
  [],
  { outFormat: oracle.OBJECT, metaData:false },
  function(err, result)
  {
    // ...
  });
```

leolmi:
4th commit: I fixed some function names to avoid problems with case

lib/oracledb.js
```diff
@@ -87,6 +87,122 @@ function extend(oracledb) {
         value: 0,
         enumerable: true
       },
+      DBTYPE_VARCHAR: {
+        value: 1001,
```
cjbj: Maybe use the OCI values?  Or 1000+the OCI value. One reason the node-oracledb type values start at 2000 is to avoid clashing with OCI type numbers

cjbj:
@leolmi to set expectations: I think we should target this for the release after Promises arrive.  We have a lot to do at the moment in that regard.

lib/oracledb.js
leolmi: I can only get at some point, beyond that I don't have the knowledge, feel free to fix everything that needs to be corrected.

cjbj: @leolmi thanks - we can take it from here.

sanfords:
This branch fixes a really important show-stopper.  And the work is _almost_ done.  Any chance you can do a priority inversion on this vs. promises?

cjbj:
@sanfords there has been a lot of demand for Promises and we are near the end of the release cycle.

cjbj:
Yay! 1.10 is out with extended metadata support.  See https://blogs.oracle.com/opal/entry/node_oracledb_1_10_has
Thanks @leolmi!

/cc @sanfords

PR #388, opened 2016-03-22
Make API compatible with Promises

jrop:
I am opening this pull request for discussion.  Theoretically, this does not break the current API, but extends it to return a Promise _and_ also call an (optionally) provided node-style callback.

## Concept

The concept is to convert a method of the form:

```js
function someAsyncFunction(callback) {
    // ... do async work, and at some point later do:
    callback(err, result);
}
```

...and converted to support both Promises _and_ callbacks:

```js
function someAsyncFunction(callback) {
    // make a callback that wraps the passed one that automatically resolves/rejects an internal promise
    callback = deferred(callback);

    // ... do async work, and at some point later do:
    callback(err, result);

    // return the promise that gets resolved by a line similar to the one above:
    return callback.promise;
}
```

## Extended Oracle API

What the commit contained in this pull request allows for is a backward compatible node-style callback API, while supporting promises.  This means that this is still valid code:

```js
oracle.getConnection(CONFIG, (err, conn) => {
	conn.execute('SELECT * FROM MY_TABLE WHERE ROWNUM < 5', { }, { resultSet: true }, (err, res) => {
		res.resultSet.getRow((err, row) => {
			console.log(row)
		})
	})
})
```

But more interestingly, the API methods also return promises, which means that when using a version of Node.js that supports generators, and also utilizing the Node module `co` (that allows generators to support promises), this code can now be written:

```js
co(function * () {
	try {
		const conn = yield oracle.getConnection(CONFIG)
		const res = yield conn.execute('SELECT * FROM CSUD_PCI WHERE ROWNUM < 5', { }, { resultSet: true })

		let row = null
		while (row = yield res.resultSet.getRow()) {
			console.log(row)
		}
		yield res.resultSet.close()
		yield conn.release()
	} catch (e) {
		// handle error and cleanup resources
	}
})
```

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -18,13 +18,15 @@
  *****************************************************************************/
 
 var resultset = require('./resultset.js');
+var deferred = require('./deferred.js');
 
 // This execute function is used to override the execute method of the Connection
 // class, which is defined in the C layer. The override allows us to do things
 // like extend out the resultSet instance prior to passing it to the caller.
-function execute(a1, a2, a3, a4) {
+function execute(a1, a2, a3) {
+  var args = deferred.variadic(arguments);
   var self = this;
-  var executeCb;
+  var executeCb = args.callback;
   var custExecuteCb;
 
   // Added this check so that node doesn't hang if no arguments are passed.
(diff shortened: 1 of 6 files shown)


dmcghan:
@jrop I've written another implementation for this type of support which can be seen here:
https://github.com/dmcghan/node-oracledb/tree/optional_promise_support

A description can be seen here:
https://jsao.io/2016/03/extending-the-javascript-layer/

Could you have a look and let me know what you think?

dmcghan:
Just a couple of thoughts... I previously added promise support to a wrapper I was working on for the driver. In that case I added promise related code inside of each function that needed promise support (all async functions). The problem with that approach is that it made each function just a little more complex than it was before. Had to be understood, maintained, etc. 

In the implementation that I pointed you to, I avoided doing that altogether. The functions only need to worry about the canonical callback API (which keeps them a bit simpler) and the util.promisify function does the heavy lifting for promises. 

That's the main difference I see between the implementations.

jrop:
Hi @dmcghan, I like the idea of `util.js:promisify` and how it detects whether a callback was passed.  @ jonathanong has a module that does the same thing called `thenify`.  I have switched to a similar approach on my local branch, but unfortunately that adds a dependency.

dmcghan:
@jrop If you agree with the approach, then I will likely use the code I showed you. Is there anything wrong with the promisify function? Why would we need to add a dependency on `thenify`?

jrop:
@dmcghan There was nothing wrong with the promise implementation; I was just thinking out loud I guess :).  I was just juggling the pros/cons of pushing off maintenance to another module (thenify).  Of course, the con is that adds a dependency.

Do you know if your work will get merged into this repo at some point?

dmcghan:
Sorry, we can't give timelines. :( Hopefully the blog post provides indication that these are high priorities.

cjbj:
@jrop it sounds like you are OK with @dmcghan's implementation?  Can you test it and let us know?  

As a rule, PR's need to follow https://github.com/oracle/node-oracledb/blob/master/CONTRIBUTING.md and have the OCA accepted.

jrop:
@cjbj @dmcghan I just tested dmcghan's branch, and it seems to be working fine!  Can't wait until this gets merged in.

Yeah, I realize that I did not sign the OCA; My main purpose in submitting this PR was to open up discussion.  If interest was shown to merge it in, I would have signed the OCA.  Now I know to read your blog for news-- didn't know this existed until now!

BTW, thanks for all the great work on this module!

cjbj:
@jrop thanks for testing. If we did an early GitHub-only release would you be interested in submitting a few functional tests to help us get it production-ready?  You'd need an OCA for that!

jrop:
@hexid You might also be interested in this.
@cjbj As long as time allows, absolutely!

cjbj:
@jrop @hexid I'm closing this PR now that @dmcghan's work is in the 1.9 development branch.  Please test and let us know what you think. See https://github.com/oracle/node-oracledb/issues/410#issuecomment-211724112 and the release notes https://blogs.oracle.com/opal/entry/node_oracledb_1_9_0

PR #393, opened 2016-03-28
Adds instantclient support

bchr02:
*I have created https://github.com/bchr02/instantclient. This new node module provides the ability to easily install instantclient directly from Oracle and this pull request adds support for using it.*

This pull requests does the following:

1. Within package.json adds instantclient as a dependency.
2. Within package.json scripts section, sets ```"install": " "``` and ```"postinstall": "node_modules/.bin/instantclient && node-gyp rebuild",```. By default when one does not include an install section it is defaulted to ```"install": "node-gyp rebuild"``` [(source)](https://docs.npmjs.com/misc/scripts#default-values). We need to run instantclient prior to node-gyp rebuild therefore I have moved it to the postinstall section *after* instantclient is ran. This is to allow the instant client files to download (if needed) as they will be needed for building.
4. modifies lib/oracledb.js so that it detects whether this instantclient folder exists and if so automatically sets the necessary environment variables.
5. modifies binding.gyp so that if the new instantclient folder is found it takes precedence. *Currently, I have only modified the Windows section. So we still need to update the Linux and Mac sections to work the same way.*

Signed-off-by: Bill Christo <billchristo@gmail.com>

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -86,8 +86,8 @@
     [
       "OS=='win'", {
         "variables" : {
-          "oci_lib_dir%": "<!(IF DEFINED OCI_LIB_DIR (echo %OCI_LIB_DIR%) ELSE (echo C:\oracle\instantclient\sdk\lib\msvc))",
-          "oci_inc_dir%": "<!(IF DEFINED OCI_INC_DIR (echo %OCI_INC_DIR%) ELSE (echo C:\oracle\instantclient\sdk\include))",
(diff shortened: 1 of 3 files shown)


cjbj:
@bchr02 After looking at this, seeing the number of submodules used that we'll need to technically & legally review - and then track to ensure security etc, my initial feeling is that this is better as a non-Oracle project.  But we'll certainly do a deeper analysis.

bchr02:
If we could directly download the files uncompressed and not need to authenticate this would cut down the modules pretty much entirely. 


    _____________________________

cjbj:
@bchr02 as we discussed, I want to keep this out of node-oracledb and look at a generic solution.

PR #404, opened 2016-04-08
Better error msg when Instant Client is missing

bchr02:
If there is a pending error and Instant Client is not found in PATH then an error message is displayed prior to `throw err;`.

Otherwise, if it is found in PATH then the directory it is **first** found in is outputed with a message asking the user to check that the correct version is installed.

Signed-off-by: Bill Christo billchristo@gmail.com

### DIFF
--- a/lib/oracledb.js
+++ b/lib/oracledb.js
@@ -48,7 +48,7 @@ try {
     fileMatch = function(dir, match){
       try {
         return fs.readdirSync(dir).some(function(file){
-          if(file.indexOf(match)>=0) {console.log(dir +'/'  +file);return true;}
+          if(file.substr(0, searchString.length) === searchString) {return true;}
         });
       }
       catch (err) {
--- a/lib/oracledb.js
+++ b/lib/oracledb.js
@@ -25,7 +25,6 @@ var connection = require('./connection.js');
 var path;
 var fs;
 var fileExists;
-var fileMatch;
 
 try {
   oracledbCLib =  require('../build/Release/oracledb');
@@ -45,21 +44,10 @@ try {
       }
     };
     
-    fileMatch = function(dir, searchString){
-      try {
-        return fs.readdirSync(dir).some(function(file){
-          if(file.substr(0, searchString.length) === searchString) {return true;}
-        });
-      }
-      catch (err) {
-          return false;
-      }
-    };
-    
     process.env.PATH.split(path.delimiter).forEach(function(dir) {
       var msg = function(){
         console.log();
-        console.log('**** Found Oracle Instant Client in PATH in the following directory: ' + dir);
+        console.log('**** Found an Oracle client library in PATH in the following directory: ' + dir);
         console.log('**** But the version is either corrupt or not for the same OS and/or Node.js processor architecture in which you\'re running on.');
         console.log('**** Your Node.js processor architecture is ' + process.arch + '.');
         console.log('**** If you have multiple Instant Client folders defined in PATH, only the first one is loaded.');
@@ -68,13 +56,14 @@ try {
         throw err;
       };
       
-      if(fileExists(path.join(dir, 'oci.dll')) || fileMatch(dir, 'libclntsh.')) {msg();}
+      if(fileExists(path.join(dir, 'oci.dll'))) {msg();}
       
     });
     
     console.log();
-    console.log('**** Oracle Instant Client directory was not found.');
-    console.log('**** Add the Instant Client directory to your PATH environment variable, then try again.');
+    console.log('****');
+    console.log('**** AN ORACLE CLIENT LIBRARY WAS NOT FOUND.');
+    console.log('****');
     console.log();
     throw err;
   }
--- a/lib/oracledb.js
+++ b/lib/oracledb.js
@@ -23,7 +23,9 @@ var Lob = require('./lob.js').Lob;
 var pool = require('./pool.js');
 var connection = require('./connection.js');
 var path;
-var existsSync;
+var fs;
+var fileExists;
+var fileMatch;
 
 try {
   oracledbCLib =  require('../build/Release/oracledb');
@@ -31,22 +33,47 @@ try {
   if (err.code === 'MODULE_NOT_FOUND') {
     oracledbCLib = require('../build/Debug/oracledb');
   } else {
-    
     path = require('path');
-    existsSync = require('fs').existsSync || require('path').existsSync;
+    fs = require('fs');
+    
+    fileExists = function(fullPathToFile) {
+      try {
+          return fs.statSync(fullPathToFile).isFile();
+      }
+      catch (err) {
+          return false;
+      }
+    };
+    
+    fileMatch = function(dir, match){
+      try {
+        return fs.readdirSync(dir).some(function(file){
+          if(file.indexOf(match)>=0) {console.log(dir +'/'  +file);return true;}
+        });
+      }
+      catch (err) {
+          return false;
+      }
+    };
     
     process.env.PATH.split(path.delimiter).forEach(function(dir) {
-      if(existsSync(path.join(dir, 'oci.dll'))) {
-		  console.log();
-		  console.log('**** Found Oracle Instant Client in PATH in the following direcory:' + dir);
-		  console.log('**** Make sure the version you have installed is the correct one.');
-		  console.log('**** Needs to match your the bitness (32bit/64bit) of your Node.js and be for the Operating System that you are using.');
-		  console.log();
-		  throw err;
-		}
+      var msg = function(){
+        console.log();
+        console.log('**** Found Oracle Instant Client in PATH in the following directory: ' + dir);
+        console.log('**** But the version is either corrupt or not for the same OS and/or Node.js processor architecture in which you\'re running on.');
+        console.log('**** Your Node.js processor architecture is ' + process.arch + '.');
+        console.log('**** If you have multiple Instant Client folders defined in PATH, only the first one is loaded.');
+        console.log('**** Therefore, the first to appear needs to be correct one.');
+        console.log();
+        throw err;
+      };
+      
+      if(fileExists(path.join(dir, 'oci.dll')) || fileMatch(dir, 'libclntsh.')) {msg();}
+      
     });
+    
     console.log();
-    console.log('**** Oracle Instant Client directory not found.');
+    console.log('**** Oracle Instant Client directory was not found.');
     console.log('**** Add the Instant Client directory to your PATH environment variable, then try again.');
     console.log();
     throw err;
--- a/lib/oracledb.js
+++ b/lib/oracledb.js
@@ -22,6 +22,8 @@ var oracledbInst;
 var Lob = require('./lob.js').Lob;
 var pool = require('./pool.js');
 var connection = require('./connection.js');
+var path;
+var existsSync;
 
 try {
   oracledbCLib =  require('../build/Release/oracledb');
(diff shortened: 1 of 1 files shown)


lib/oracledb.js
```diff
@@ -22,13 +22,33 @@ var oracledbInst;
...
     oracledbCLib = require('../build/Debug/oracledb');
   } else {
+    
+    path = require('path');
+    existsSync = require('fs').existsSync || require('path').existsSync;
+    
+    process.env.PATH.split(path.delimiter).forEach(function(dir) {
+      if(existsSync(path.join(dir, 'oci.dll'))) {
+		  console.log();
+		  console.log('**** Found Oracle Instant Client in PATH in the following direcory:' + dir);
```
cjbj: Good idea to improve this.  Can the messages cater for users with ORACLE_HOME installs too?  Something like "Found the Oracle client library ..."  Testing PATH will only work on Windows; I'd be OK making the improved validation work on Windows-only for the moment.

bchr02: Yes, I will modify to cater to ORACLE_HOME too. Good idea. Also, for Linux we could search for libclntsh.so.12.1 or libclntsh.so.11.1, right?

cjbj: For non-Windows I think you'd be OK checking for 'libclntsh.*'  Harder to portably check the link has been created.

sagiegurari:
I recommend not to use console.log for valid things, instead to use https://www.npmjs.com/package/debuglog
Otherwise users will always see messages without a way to disable them.

bchr02:
@sagiegurari isn't it kinda of a moot point since we are throwing an error. But while we're on the topic shouldn't we be handling the errors asynchronously by passing the error up the callback chain instead of try/catching it?

sagiegurari:
error you say? you have printouts to console.log for valid scenarios.
so either remove it, or use debuglog

````js
 if(existsSync(path.join(dir, 'oci.dll'))) {
 		  console.log();
 		  console.log('**** Found Oracle Instant Client in PATH in the following direcory:' + dir);
````

**Comment by bchr02** [comment]
It's not a valid scenario. This only prints out when it's the wrong instant client version.

**Comment by sagiegurari** [comment]
ok, now i get you. than forget it, you are right.

**Comment by sagiegurari** [comment]
in windows, i saw an issue for people who used node 64 bit with oci 32 bit.
Well... they had multiple ocis (don't ask me why) and it compiled with one and ran with another (again don't ask me why, wasn't my pc but people have a mess) and I had to use dependency walker to see that issue.
I wonder if you can provide that info as well, meaning provide the node 32/64 bit version and the oci 32/64 bit version (if you find the oci).
what do you think?

**Comment by dmcghan** [comment]
Not all errors are asynchronous:
https://www.joyent.com/developers/node/design/errors

This try/catch is only executed once to bring in the C lib when the module is first "required" in. There is no callback chain in this context anyway, so throwing is the correct behavior.

**Comment by bchr02** [comment]
@dmcghan okay that makes sense then. 

@sagiegurari yea i thought about including the node.js version but didn't know how to extrapolate the same information from the dll, so for now I decided to hold off. As far as having multiple instant clients in path, the first one to appear is the one that is used which is why only the first one's directory is displayed. Perhaps the error message should indicate this.

File: lib/oracledb.js (hunk 1 above)
Comment (dmcghan) [inline lib/oracledb.js]: This doesn't look right:
```javascript
    path = require('path');
    existsSync = require('fs').existsSync || require('path').existsSync;
```

I can't find `existsSync` in `path`. Also, existsSync is deprecated:
https://nodejs.org/api/fs.html#fs_fs_existssync_path

This the main answer here offers an alternative:
http://stackoverflow.com/questions/4482686/check-synchronously-if-file-directory-exists-in-node-js

bchr02: Yes, you're right. I will update it. 

- sent from iPhone.

sagiegurari:
maybe this can help:
http://stackoverflow.com/questions/495244/how-can-i-test-a-windows-dll-to-determine-if-it-is-32bit-or-64bit
or
http://stackoverflow.com/questions/197951/how-can-i-determine-for-which-platform-an-executable-is-compiled

These are not js implementations but i'm sure you can convert it to js to validate.
This means reading the dll in certain area and checking for the value in the header if it is 32 or 64 bit.
Didn't find an npm module written for that.
I'm not sure about linux or mac.

bchr02:
I was just doing some tests on Windows. I didn't have any environment variables set. I opened a command prompt and typed ```set ORACLE_HOME=c:\oracle\instantclient``` but oracledb would not work. I had to set the directory in PATH for it to work. Why is that? It's my understanding that ORACLE_HOME takes president over PATH.

bchr02:
In response to everyone's feedback, I just pushed the following changes:
1. now using statSync to check whether oci.dll exists
2. added fileMatch function to allow the searching of any files beginning with ```libclntsh.```
3. added more information to the message that's displayed when an incorrect lib is detected.

I haven't added the scanning of ORACLE_HOME yet. I am waiting to get some feedback from @cjbj on this. Need to better understand when ORACLE_HOME is used and whether it is preferred over PATH. As explained, on Windows having it defined does not appear to influence anything. Maybe it's only for Linux?

kubo:
> It's my understanding that ORACLE_HOME takes precedent over PATH.

As for instant client, ORACLE_HOME is referred only when TNS_ADMIN isn't set and it is used to locate configuration files such as tnsnames.ora or sqlnet.ora.

As for full client, ORACLE_HOME must be set. Otherwise Oracle client doesn't work. On Unix ORACLE_HOME must be set as an environment variable. On Windows it is set in the Windows registry.

kubo:
>  added fileMatch function to allow the searching of any files beginning with libclntsh.

It doesn't work on Unix because the environment variable PATH is not used to locate dependent shared libraries on Unix.

On Linux, the location of libclntsh.so is embedded in the compiled node-oracledb binary by the linker option `-Wl,-rpath,/directory/containing/libclntsh`.
See https://github.com/oracle/node-oracledb/blob/v1.8.0/binding.gyp#L30
If libclntsh.so is not found in the embedded path, the dynamic loader tries to find libclntsh.so in the environment variable LD_LIBRARY_PATH, directories in `/etc/ld.so.conf` (`/etc/ld.so.cache` is created from this file by the ldconfig command.),  `/lib` and `/usr/lib`.
See http://man7.org/linux/man-pages/man8/ld.so.8.html

On other Unix platforms, similar but different ways are used to locate libclntsh.so.

bchr02:
Thanks @kubo 

>As for instant client, ORACLE_HOME is referred only when TNS_ADMIN isn't set and it is used to locate configuration files such as tnsnames.ora or sqlnet.ora.

In my testing on Windows, I did not have TNS_ADMIN set and I can confirm that although I had ORACLE_HOME set it was not used. So I'm still not sure why that is. 

In regards to your second reply, based on that I think I may need to make the first section, the one that searches for the library, only for Windows and for everything else just have a standard message explaining that the library was not found. Or perhaps I can search the LD_LIBRARY_PATH?

Before I change anything again, I will wait a bit to give everyone the opportunity to chime in.

kubo:
> In my testing on Windows, I did not have TNS_ADMIN set and I can confirm that although I had ORACLE_HOME set it was not used. So I'm still not sure why that is. 

ORACLE_HOME is unrelated to the location of OCI.DLL. It is used only to locate network configuration files when TNS_ADMIN is not set in instant client. If you don't use network configuration files such as tnsnames.ora, ORACLE_HOME has no effect.

> In regards to your second reply, based on that I think I may need to make the first section, the one that searches for the library, only for Windows and for everything else just have a standard message explaining that the library was not found. Or perhaps I can search the LD_LIBRARY_PATH?

+1 to the former.
It is hard to implement same feature for Unix.

cjbj:
@bchr02 to surface my [earlier comments](https://github.com/oracle/node-oracledb/pull/404/files/6c29e2e8dbbe2b1363aa809d640631a657426033#r59102730) I think it will easier to make the change (for this PR, at least) specific to Windows. The general message terminology needs to use 'Oracle client' instead of 'Oracle Instant Client' since PATH may be using an Oracle Home-style install and not Instant Client.

bchr02:
>I think it will easier to make the change (for this PR, at least) specific to Windows. The general message terminology needs to use 'Oracle client' instead of 'Oracle Instant Client' since PATH may be using an Oracle Home-style install and not Instant Client.

@cjbj Okay, I just pushed those changes. Thanks.

cjbj:
@bchr02 in 1.11 we emit a new NJS error message when the node-oracledb binary can't be loaded but we haven't added any of the investigatory heuristics of this PR.

cjbj:
@bchr02 where do you want to go with this?  I adopted the ideas into https://github.com/oracle/node-oracledb/blob/dev-2.0/lib/oracledb.js (thank you!) without the port specific Windows check for oci.dll.  Do you want to rework the PR and add it under a platform check for Windows?  It should probably mention needing the VS Redistributable.  There could be some smarts like looking for a BASIC_README or BASIC_LITE_README to get a version number and then printing out which VS Redistributable is needed?

bchr02:
@cjbj I would love to. I’ll start working on something. 

So, what did you have in mind by looking at the readme’s. Is there a version number within them, some metadata on the files themselves or does the mere fact that they exist mean that VS 2010 is needed over 2005?

cjbj:
@bchr02 If your code finds an oci.dll and there is a BASIC_README or BASIC_LITE_README then you know the user has Instant Client - and the user must separately install a redistributable. You can parse the version out of the readme from the `Client Shared Library 64-bit - 12.2.0.1.0` line and then print (from a hardcoded list) which VS redistributable is needed.  Or run`dumpbin /headers oci.dll` and have a lookup table from that number to the VS Redist (see https://github.com/oracle/node-oracledb/blob/v2.0.14-dev/INSTALL.md#--366-install-the-visual-studio-redistributables) - that might be easier now I think of it.

Some of the logic could also be added into https://github.com/oracle/node-oracledb/blob/v2.0.14-dev/package/oracledbinstall.js#L336

These are just brainstorm ideas from a non-windows user.

kubo:
@cjbj Is the getInfo() function in [oracledb.js](https://github.com/oracle/node-oracledb/blob/dev-2.0/lib/oracledb.js) called when OCL.DLL or VS redistributable is missing?

getInfo() is called only when `err.code === 'MODULE_NOT_FOUND'`. However, as far as I checked node.js source code (Well, I have not checked it by running node.js. Sorry if I mistake), it is set only when a required file itself ("oracledb.node" in this case) is not found.

cjbj:
@kubo Now that ODPI-C uses dlopen() it will pick up some errors first so you're right: the [getInfo()](https://github.com/oracle/node-oracledb/blob/dev-2.0/lib/oracledb.js#L76)  call may need to be higher in oracledb.js.  The intention was for it to be  called when the add-on library (oracledb.node) exists but can't be loaded.  I'll do some tweaking.

cjbj:
@kubo.  Something like https://gist.github.com/cjbj/78a587855a81edec3fb4dedc8aece522 ?

cjbj:
I updated the gist.

kubo:
@cjbj I checked the gist on Linux. It looks good.

bchr02:
@cjbj In light of https://github.com/oracle/odpi/issues/48  which route are we going with this? will this be done in ODPI-C or oracledb.js?

cjbj:
@bchr02 I think Anthony will make some changes.  Give it a day or two to see what he commits.

cjbj:
The v2.0.15 messages might contain too much information; for example if there is no oracledb.node binary, the messages mention this and then talk about Instant Client.  I expect some readers will focus on the wrong parts of the message.

cjbj:
@bchr02 is it time to close this?

bchr02:
Yes, thank you.

PR #470, opened 2016-07-06
enchancement-469: NJS-037 Now has extra details about bind IN parameter.

### LINKED ISSUE #469: More details for NJS-037.
I am invoking a PLSQL procedure which takes a bunch of inputs. Many of them are arrays of STRING, NUMBER and DATE types.

I'm creating a binding for these values before sending them to the procedure. I'm following the examples given here - https://github.com/oracle/node-oracledb/blob/master/examples/plsqlarray.js

I get an NJS-037 error. The message I get is this `incompatible type of value provided`.

From this, all I can conclude is that I'm sending an incompatible value for a certain type. But I don't know for which one. 

Is there a way to get the details? Or a debug flag that I can enable to get the sql that is finally executed?

hariprasadkulkarni:
This fixes #469 . Adding extra information about which bind parameter incompatible types.

The assumption is that `bind->key` is not null (as the other members of the structure are already defined and being used).

Signed-off-by: Hariprasad Kulkarni kulkarni.hariprasad@gmail.com


cjbj:
Hi @hariprasadkulkarni Thanks for this! Did you submit your OCA paperwork?

hariprasadkulkarni:
Not yet, I will do it ASAP.

hariprasadkulkarni:
@cjbj I've mailed the paperwork.

pvenkatraman:
This context is common for both bind-by-name and bind-by-position.  In case of bind-by-position, bind->key will be empty string and will not help.  Can you please add another message and use that message in that case providing the bind-position.  Rephrase messages accordingly. Also can you add couple test cases to hit these scenarios. Thx

hariprasadkulkarni:
In src/njs/src/njsConnection.cpp, `GetBinds` is an overloaded function.

For "Object" (Bind by name), it populates the `key` parameter.
For "Array" (Bind by position), it does not populate the `key` parameter. And, as far as I can see, it does not store the `index` anywhere in the `Bind` struct. The `index` is lost here.

Are you OK with adding another parameter to `Bind`?

pvenkatraman:
Can you try the following:
std::vector<Bind*>::iterator it;
it = std::find ( binds.begin(), binds.end (), bind );
if ( it == binds.end () )
{
	// not found
}
else
{
	int index = std::distance ( binds.begin(), it );
}

We should always find the Bind structure from the vector.
Thanks,

hariprasadkulkarni:
@pvenkatraman Sorry, my notifications on github were disabled.

So the error `errIncompatibleTypeArrayBind` is raised in `GetInBindParamsArray`. In that function, I have a single bind and not the bind vector to get the index.

Furthermore, this error is not bubbled up, so that we can use it in the function where we have the bind vector.

pvenkatraman:
@hariprasadkulkarni  Thanks for your response.  Connection::GetInBindParamsArray () gets executeBaton as parameter, and executeBaton->binds is the vector of Bind structures.  My suggestion to find the index from the vector collection itself is to minimise adding more fields into structure which may never get used. Thanks,

hariprasadkulkarni:
@pvenkatraman I tested it on my system. I'm getting the errors. I have to look into the test suite to add some tests.

hariprasadkulkarni:
@pvenkatraman @cjbj Any comments on this?

cjbj:
@hariprasadkulkarni I admit we've been working on other things for node-oracledb.  The last status on this was that you were going to add tests - are you still planning to do that?

hariprasadkulkarni:
I haven't looked into adding tests for this. Let me check it out.

cjbj:
@hariprasadkulkarni if you want, we can take it from here and do what is needed.

hariprasadkulkarni:
@cjbj That would be great!

cjbj:
@hariprasadkulkarni thanks for this PR.  @pvenkatraman tidied it up, tests were created, and it has been merged to 1.13.

hariprasadkulkarni:
@cjbj Thank you!

Issue #529, opened 2016-10-14
Readme-Example does not close its connection


stealthybox:
#### 1. What error(s) you are seeing?

After basing application code off of the example code in the README, our team experienced leaked connections to our database.
#### 2. What caused the problem?

The example code does not demonstrate how to explicitly close connections.
When running queries intermittently in a loop, large numbers of connections can build up over time.

cjbj:
We had discussions about this and so far I've won - but that may change :)  The README example is a simple 'this is what it looks like' demo.  The https://github.com/oracle/node-oracledb/tree/master/examples are the demos to follow for practical examples.

stealthybox:
@cjbj oops, I just submitted a PR I was staging
-- sorry I didn't see your comment
(I was signing my OCA)

Could we throw in a note that points people in the right direction then?
My use case was very simple so I grabbed that code snippet assuming it would be compliant with the practices in the examples directory.

Our devs aren't familiar with this driver yet, so leaky code made it into our build pipeline.
Thanks for the quick response!

sagiegurari:
Have to agree with @cjbj 
The examples also won't show how you get the connection in the first place as well.
If you go to that direction there are just too many permutations and the examples will be a real mess that no one will understand.
For example others will complain why after every execute they need to release the connection and can't reuse it, and why? Because the example had a release after execute.
Developers need to read the basics before coding.

bchr02:
@cjbj @sagiegurari I have to disagree. The section is titled "A simple query example with callbacks" and not "An incomplete example of what the API looks like". I think the current title is fine, but it should include something as basic and necessary as closing the connection. Even a basic query needs to close its connection, right? Leaving it is clearly misleading, otherwise we would't have this issue. Anyways that's my thoughts. 😃

@sagiegurari Where you write:
> The examples also won't show how you get the connection in the first place as well.

I think you are incorrect with that statement because the example does show this.

>If you go to that direction there are just too many permutations and the examples will be a real mess that no one will understand.

There is currently only one example on the main README's Example section which is the subject of this issue. So, I don't think there is an issue of "too many permutations". Even if additional example are added they will have a sub section heading.

>others will complain why after every execute they need to release the connection and can't reuse it, and why? Because the example had a release after execute.

By this logic we shouldn't have any example because people will think whatever they want to think.

> Developers need to read the basics before coding.

I agree, but if someone is just dabbling at using node with oracledb and they see a section titled "A simple query example..." one could argue that this is the basics.

[...]

PR #530, opened 2016-10-14
docs-529 Updated example code to release the connection

stealthybox:
This addresses #529.

Signed-off-by: leigh schrandt leigh@null.net


cjbj:
Your OCA was approved.  We'll pick up this fix as discussed in #529

cjbj:
I used this as the basis for changes in 1.12.0-dev.  Thanks for the PR!

Issue #818, opened 2017-12-28
Uncaught Error: NJS-045: cannot load the oracledb add-on binary for Node.js 7.9.0


kwent:
Hi,

So i spent a fair amount of time trying to make my electron app works. I followed the installation guide with Zip File method but i hit the issue below which prevent the all app to start.

```
Uncaught Error: NJS-045: cannot load the oracledb add-on binary for Node.js 7.9.0 (linux, x64)
Cannot load /home/osboxes/app/release/linux-unpacked/resources/app.asar/node_modules/oracledb/build/Release/oracledb.node
The module '/tmp/.org.chromium.Chromium.QS5Iaa'
was compiled against a different Node.js version using
NODE_MODULE_VERSION 57. This version of Node.js requires
NODE_MODULE_VERSION 54. Please try re-compiling or re-installing
the module (for instance, using `npm rebuild` or`npm install`).
Node-oracledb installation instructions: https://github.com/oracle/node-oracledb/blob/master/INSTALL.md
You must have 64-bit Oracle client libraries in LD_LIBRARY_PATH, or configured with ldconfig.
If you do not have Oracle Database on this computer, then install the Instant Client Basic or Basic Light package from 
http://www.oracle.com/technetwork/topics/linuxx86-64soft-092277.html
```

The system node version used is v8.2.0. So i have no idea why the error message is mentioning v7.9.0 ?

I installed nvm and switch to 7.9.0 version and same issue.

Not sure about NODE_MODULE_VERSION message as well. 54 is referencing to any nodejs version when i take a look at the official page: https://nodejs.org/en/download/releases/

I tried to rebuild extension but nothing i get stuck with this message. Any idea ?

Regards

dmcghan:
@kwent This line "The module '/tmp/.org.chromium.Chromium.QS5Iaa'..." seems suspect.

What version of Electron are you running?

What version of the driver (node-oracledb) are you using?

What do you get if you run: `console.log(process.versions)`?

Sorry, it's been a while, but when do dependencies of your Electron app get installed/compiled?

You may need to pick your Node.js version, then install Electron, then install the driver. After that, don't change your Node.js version unless you want to reinstall Electron and the driver.

kwent:
@dmcghan  Here is a demo repo i can repro the issue: https://github.com/kwent/electron-react-boilerplate

You can download `.deb` artifact at : https://circleci.com/gh/kwent/electron-react-boilerplate/2#artifacts/containers/0

[...]

PR #851, opened 2018-02-16
Use true-case-path to fix re-entrancy issue on Windows and Webpack

amilajack:
This issue is a fix for https://github.com/oracle/node-oracledb/issues/818. This issue is similar to https://github.com/sass/node-sass/pull/2149, which

> When using node-sass through multiple entry points (as can happen when using webpack) on Windows, the casing of the drive letter in the path can be inconsistent. Since Node is case-sensitive to a fault, this results in Node attempting to create a second instance of the native binding, which results in a Module did not self-register error.
>
> The fix is to use true-case-path to normalize the binding path, to prevent unintended duplicate bindings from occurring.

### DIFF
--- a/lib/oracledb.js
+++ b/lib/oracledb.js
@@ -39,7 +39,13 @@ var binaryDebugPath = '../build/Debug/oracledb.node';
 // --debug oracledb' was used.  Typically only the maintainers of
 // node-oracledb do this.
 try {
-  oracledbCLib =  require(binaryReleasePath);
+  if (process.platform === 'win32') {
+    var trueCasePath = require('true-case-path');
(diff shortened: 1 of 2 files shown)


cjbj:
I appreciate you doing this bug

- you didn't sign the OCA before submitting

- I really don't want 3rd party code included for legal & security reasons.

amilajack:
I have followed the directions [here](http://www.oracle.com/technetwork/community/oca-486395.html) for the OCA. I submitted mine a few minutes ago. Does it take time to process?

cjbj:
I've no plans to merge this because it injects 3rd party code into the install - a security issue.  I'd be happy to merge a PR that directly patched node-oracledb code.

PR #891, opened 2018-04-14
Added Support for STAGING packages in MAKEPKG.BAT

danilohgds:
Things to notice: 

Replacing the strings in oracledbinstall.js with a FOR LOOP in .bat would cause an issue, so SED was used. This means cygwin installation is needed.

Signed-off-by: Danilo Henrique Garcia da Silva <danilo_hgds@hotmail.com>


cjbj:
@danilohgds Excellent. I'll let the OCA team know to look out for it when they do their next batch of processing.

danilohgds:
@cjbj  I had to resend it on monday, section 7 was missing a X mark...probably they will see it today I guess.

danilohgds:
@cjbj , I just noticed that. I was doing some tests, forgot to switch branches my bad.

The form was missing my house number, it is somewhat messy. I will resend soon, had a tough week.

cjbj:
@danilohgds did you send a new OCA application?

danilohgds:
@cjbj many days ago, haven't heard anything since.

danilohgds:
@cjbj my OCA got accepted, just received back feedback, it should show up in the contributors list in some days.

cjbj:
@danilohgds congrats!  I got the email too but haven't worked down my inbox yet.  We'll take a quick look at the MR, maybe suggest one or two tweaks, and then copy the changes to our internal repo ready for the next code push to GitHub - I was thinking of pushing the current 2.3-dev code but not doing a binary release until features have landed and been tested.

pvenkatraman:
@danilohgds  - looks like you are using UNIX utility SED on Windows.  What version of SED you are using and which publisher?  Is -i option supported by all versions of SED? Thanks

danilohgds:
Hi @pvenkatraman , I`m using cygwin to achieve the SED functionality in windows. Cygwin DLL is 2.10.0.

The original FOR loop with keys would not handle the file correctly in the bat file so I changed it.

cjbj:
@pvenkatraman I know you would like some refactoring to avoid the small amount of duplication, and I'd prefer some alignment with the Linux makefile targets, but I'm OK with us merging this PR as it is so we can get on with the 2.3 release.  The code works for people who need it and can be be refactored later if it becomes an issue.

pvenkatraman:
@danilohgds instead of using "sed" utility can you use findstr.exe.  SED is not expected to be available in all Windows systems and looks like cygwin and MKSNT versions vary in command line options.  FINDSTR.EXE is a standard windows utility and is expected to be available on all Windows systems.

danilohgds:
I will switch to findstr then, should habe it working till friday I hope.

cjbj:
@danilohgds if you're spending time on it, can you also look at removing the duplication of creating the binary - see how Makefile separates the binary & generic package builds?

danilohgds:
@cjbj So far I haven't been able to workaround the bug involving the FOR tokens syntax.

In my opinion findstr will not help me because the problem is within the For/Token logic itself..

```

Set "hostname=%NODE_ORACLEDB_PACKAGE_HOSTNAME%"
Set "pathname=%NODE_ORACLEDB_PACKAGE_URL_PATH%"
set "searchpath=/oracle/node-oracledb/releases/download/"
set "search=github.com"

set oracleInstallFile=./oracledbinstall.js
set tempFile=./oracledbinstall.tmp.js

if exist "%tempFile%" del /f /q "%tempFile%"

for /F delims^=^ eol^= %%a in (%oracleInstallFile%) do (
   set newline=%%a
   echo !newline! >> %tempFile%
)
```

That loop alone, will cause some lines from the original JS file to go bad, for some reason which I cannot figure out.

From : 

```
if (!proxy.startsWith('http://') && !proxy.startsWith('https://')) {
      proxy = 'https://' + proxy;
    }
```
to ( note how the entire proxy.startsWith part is gone.. )

```
  if (proxy) { 
    proxyConfig.useProxy = true; 
    if (//') && //')) { 
      proxy = 'https://' + proxy; 
    } 
```

Once I figure out what is causing this, I will update the PR.

cjbj:
@danilohgds any news on this?

danilohgds:
@cjbj got the dedup removed, but I will wait based on what we discussed about the PR.

cjbj:
The 2.3 code base now contains this change.  Thank you @danilohgds.

PR #1043, opened 2019-02-04
Add query config object support

cemremengu:
I would like to add [query config object](https://node-postgres.com/features/queries#query-config-object)  support to the driver (this is available in mysql drivers as well). It is very helpful to pass a query object in various situations. (e.g. something like [this](https://github.com/nearform/sql))

I couldn't see it in the code base but hopefully it may already be available.

ps: I am not sure if this is breaking any of the tests since it is very difficult for me to setup the test environment unfortunately.

Signed-off-by: cemremengu <cemremengu@gmail.com>

### DIFF
--- a/examples/example.js
+++ b/examples/example.js
@@ -98,7 +98,8 @@ async function run() {
       // fetchArraySize: 100       // internal buffer allocation size for tuning
     };
 
-    result = await connection.execute(sql, binds, options);
+    // You can also use a query config object to query the data
+    result = await connection.execute({ sql, binds }, options);
 
     console.log("Column metadata: ", result.metaData);
     console.log("Query results: ");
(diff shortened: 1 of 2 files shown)


cjbj:
@cemremengu to drill into the real value of the PR, in the case of nearformsql, is it really too difficult to pass sql.sql and sql.binds (or sql.values?) as separate arguments to execute()?

I'd like to keep a high bar on enhancements, so we don't end up with lots of confusing alternatives (and extra doc and testing and maintenance).  Can you think of other use cases?

cemremengu:
@cjbj Actually, I think it is better to just create a wrapper around it for such particular case. Thanks for pushing me to think a little more!

PR #1460, opened 2022-02-23
Update README.md

Kaustubh01:
```
Signed-off-by: Your Name <you@example.org>
```


cjbj:
Thanks for the PR.  In general you need to sign the [OCA](https://github.com/oracle/node-oracledb/blob/main/CONTRIBUTING.md).  Are you already on the OCA list?  

Regarding the PR, it would be good to have something about Node.js, but your text looks very close to text from the Node.js pages.

Issue #465, opened 2016-06-30
capture stack trace for errors


mxriverlynn:
I'm working on a multi-layered, transactional save for an object graph with about three levels of objects in it.

As I'm building this out, I get errors because I have the wrong data type in my bindvars for the stored procs, or because I misspelled a var in the bindvars, etc.

When this happens, I get an error message, such as:

```
Error: NJS-011: encountered bind value and type mismatch in parameter 2
    at Error (native)
POST /api/entries/save 500 192.169 ms - 107
```

While this error message does tell me that I have a problem with my parameter binding, and it does tell me which parameter, it doesn't tell me which of the many saves in my code is having the problem.

When this happens, I have to debug into my code and hope I put breakpoints in the right places... or use console.log statements in my `if (err){ ... }` code to try and figure out which stored proc call is failing.

It takes a while to find the problem and fix it, because of this. And when I have a second error in my bindings, but it produces the same error message, I don't know if I solved the original one or not because I can't see where the problem originates.

It's very frustrating and time consuming.

fortunately, there's a fix for this - one that should be integrated into the library: `Error.captureStackTrace`

By adjusting my own code, I can work around the problem of not knowing where the problem occurred.

For example:

```
    connection.execute(..., function(err, result){
      if (err) { 
        Error.captureStackTrace(err);
        // ... do something with the error 
      }
  });
```

now, when my error handling code captures this, and I look at `err.stack`, I see something more like this:

```
Error: NJS-011: encountered bind value and type mismatch in parameter 2
    at /Users/derickbailey/dev/glmetadata/lib/glentries/entry/persistence.js:108:15
    at /Users/derickbailey/dev/glmetadata/node_modules/ocarina/lib/storedProc.js:30:23
    at custExecuteCb (/Users/derickbailey/dev/glmetadata/node_modules/oracledb/lib/connection.js:98:7)
```

line 108 of the persistence.js file is where I called `Error.captureStackTrace`. And while this in itself isn't ideal, it gives me the visibility I need to see where the error is, in my code. It greatly reduces the amount of time needed to find the problem, fix it, and know that I fixed this problem when another error pops up.

**Please add the use of `Error.captureStackTrace` to the core library, anywhere an error is created and returned.**

It will be a huge improvement in debugging and working with this library.

cjbj:
Good idea but is `Error.captureStackTrace(err)` the solution for the add-on?  If it is used in node-oracledb, it doesn't report the application script line.

[...]

PR #1467, opened 2022-03-14
Capture stack trace for error

sla100:
Solution for https://github.com/oracle/node-oracledb/issues/465

```
```

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -174,8 +174,8 @@ function callbackify(func) {
     // if last argument is not a function, simply invoke the function as usual
     // and a promise will be returned
     if (typeof arguments[arguments.length - 1] !== 'function') {
-      return func.apply(this, arguments).catch(function(e) {
-        Error.captureStackTrace(e);
+      return func.apply(this, arguments).catch(function st(e) {
+        Error.captureStackTrace(e, st);
         throw e;
       });
     }
--- a/lib/util.js
+++ b/lib/util.js
@@ -163,6 +163,8 @@ function checkArgCount(args, minArgCount, maxArgCount) {
 
 module.exports.checkArgCount = checkArgCount;
 
+// This expression captures the prologue of the exception also with --enable-source-maps
+const stacktraceRe = /^[\s\S]*Error: #/;
 
 // The callbackify function is used to wrap async methods to add optional
 // callback support. If the last parameter passed to a method is a function,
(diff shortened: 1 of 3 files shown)


cjbj:
@sla100 when I did a (very) quick test yesterday, moving the `new Error()` call after catch still showed the failing point of code when there was a connection error, even if other layers weren't shown.  

```
      return func.apply(this, arguments).catch(function(e){
        const context = new Error('#');
        e.stack += context.stack.replace(stacktraceRe, '');
        throw e;
      });
```

gave:

```
cjones@mac:~/n$ node t.js
Error: ORA-01017: invalid username/password; logon denied
    at /Users/cjones/n/node-oracledb/lib/util.js:182:25
    at async run (/Users/cjones/n/t.js:25:18) {
  errorNum: 1017,
  offset: 0
}
```

Which correctly identified line 25 of t.js as having my `getConnection()` call.

I only did the one test....

But if this is valid, would it be just as good a solution for identifying where the error was?  One thing that concerned me when I looked at https://github.com/oracle/node-oracledb/issues/465 was the overhead of calling `new Error()` for each node-oracledb API call.

I'd be tempted to strip out the util.js from the stack

sla100:
It seems to me that since Node 16 which using V8-7 the situation has improved:
https://v8.dev/blog/v8-release-73#async-stack-traces
https://nodejs.org/en/download/releases/

When we limit ourselves to async / await the change becomes very light. I changed my code.

cjbj:
@sla100 I'd be fine with support just for async/await.

From another quick test, what if you name the function so its stack can be omitted?:
```
      return func.apply(this, arguments).catch(function st(e) {
        Error.captureStackTrace(e, st);
        throw e;
      });
```
With this, a connection error is like:
```
Error: ORA-01017: invalid username/password; logon denied
    at async run (/Users/cjones/n/t.js:25:18) {
  errorNum: 1017,
  offset: 0
```

cjbj:
@sla100 any news on your OCA? Searching for your names in the system isn't showing me anything.

sla100:
Nope. I send email to opensource_ww_grp@oracle.com and no answer.

abrar-khan007:
Issue yet exists , seems returns bool  ,how can I check

sla100:
> @sla100 any news on your OCA?

I received a reply from the person responsible for the declarations. There is hope :)

cjbj:
> > @sla100 any news on your OCA?
> 
> I received a reply from the person responsible for the declarations. There is hope :)

It was completed: welcome!  Now I just need to find time to work with your PRs.

cjbj:
@sla100 your PRs are being looked at by @pvenkatraman .  We will merge the changes to our internal repo and then sync back to GitHub.  So don't be surprised if you don't see a normal merge on the PRs.

cjbj:
This got merged as https://github.com/oracle/node-oracledb/commit/ca165e95b8c9e5ddbaf428550435b361d7b0a0a5
Thank you!

PR #1499, opened 2022-05-26
Add link to git changes report

sosoba:
```
```


cjbj:
Good idea.  I'll make this change (and do the same for older releases in CHANGELOG) in our internal repo and then push out to GitHub with my next batch of updates.  Thanks for the suggestion.

PR #1559, opened 2023-05-28
Unification nodejs imports

sla100:
Hello. Congratulations on working on the new "thin" version!. 
I allowed myself a PR unifying the way of importing dependencies from nodejs.
Currently, the design uses a mixed style. Sometimes these are references to global objects, sometimes imports. 
IMHO there are preferred:
- explicit imports to Node resouces (Buffer, process os, etc.),
- implicit imports of ESM globals ([FinalizationRegistry](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry))

Best regards.
```
```


cjbj:
Thanks @sla100 !

Some of our new code has been itching @anthony-tuininga for a while and he is about to commit a big cleanup and simplification.  I'll let him comment on your PR and how it can be integrated.

anthony-tuininga:
Thanks, @sla100. Unless you have a good reason to do otherwise, I think I'll use this approach instead:

```javascript
const { Buffer } = require('buffer');
```

I'll make sure this gets integrated with the big cleanup. Thanks again!

sla100:
@anthony-tuininga
> I think I'll use this approach instead

Sure. It is closer to esm `import` statement.

I can also suggest some modification for `.eslintrc`. `Node` option are heavy and allow to use many global references.
```json
"env": {
  "node": true,
}
````
After being replaced by:
```json
"env": {
  "shared-node-browser": true,
  "commonjs": true,
}
```
allow only references accomplished to ESM standards and cjs require / exports.

cjbj:
@sla100 changes landed in https://github.com/oracle/node-oracledb/commit/15f10c150919bbbc36357b27a5da0c43a5d92f2f and https://github.com/oracle/node-oracledb/commit/ae623d1e8c9bf637966396710d4b3f30c4cef3e0. Thanks for the advice and contribution.

[For other readers not familiar with our PR process, we take the code from PRs into a private repo (which we need in order to work on unreleased Oracle DB feature support) and then the PR changes get committed back to GitHub.]

PR #1572, opened 2023-06-09
feature: Support bigint parameters in thin mode

sla100:
Hello. There is a simple change which allow to pass build-in bigint type to parameters in thin node.
This allows you to pass large integers without conversion to string in js and CAST on the Oracle side.
Ex.:
```js
const {rows} = await conn.execute( 'SELECT null FROM DUAL WHERE :A = (CAST :B AS NUMBER)', {
  A: 1_000_000_000_000_001n,
  B: 1_000_000_000_000_001n.toString(),
});
equal(rows.length, 1);
```

```
Signed-off-by: Sławomir Osoba <you@example.org>
```

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -195,6 +195,7 @@ class Connection extends EventEmitter {
       typeof value === 'number' ||
       typeof value === 'string' ||
       typeof value === 'boolean' ||
+      typeof value === 'bigint' ||
       Array.isArray(value) ||
       Buffer.isBuffer(value) ||
       util.isDate(value) ||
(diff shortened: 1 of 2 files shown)


sharadraju:
We would also need to think about how to retrieve data from Oracle Database as `bigint` datatype as well.

sla100:
> We would also need to think about how to retrieve data from Oracle Database as `bigint` datatype as well.

This translation is already possible for columns.

```js
const fetchTypeHandler = ({dbType, precision, scale}) => {
  if (dbType === oracledb.DB_TYPE_NUMBER && scale === 0 && precision > 15) {
    return {
      type: oracledb.STRING,
      converter: (v) => v === null ? null : BigInt(v),
    };
  }
};
```

sharadraju:
@sla100 This has been fixed in the 6.5 release. Thank you for your contribution!

PR #1766, opened 2026-01-20
value instanceof Date

denizyts:
```
Signed-off-by: Deniz Yetiş <denizyetis56@gmail.com>
```

(node:93573) [DEP0047] DeprecationWarning: The `util.isDate` API is deprecated.

This deprecation message annoys me a lot.  I hope in the future versions it will be fixed or my pull request will be merged.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -259,7 +259,7 @@ class Connection extends EventEmitter {
       Array.isArray(value) ||
       nodbUtil.isVectorValue(value) ||
       Buffer.isBuffer(value) ||
-      util.types.isDate(value) ||
+      value instanceof Date ||
       value instanceof Lob ||
       value instanceof ResultSet ||
       value instanceof BaseDbObject


barisaydinayy:
this deprecation warning annoys me a lot also ++++++++++

sharadraju:
This deprecation is done only in Node.js 25 and above, I think.
So, it is only a recent change in Node.js as it is.
What is the Node.js version that you are using?

denizyts:
> This deprecation is done only in Node.js 25 and above, I think. So, it is only a recent change in Node.js as it is. What is the Node.js version that you are using?

v22.12.0

deprecation gone after fix

sharadraju:
Thanks @denizyts. We will check this internally and make the change in the next release, if required (with due credits given to you :) )

sharadraju:
@denizyts The strange thing is, I am not seeing this warning and I am also using Node.js 22.

Can you give me an example, where you are seeing this warning? Are you using node-oracledb as part of any framework?

sudarshan12s:
Yes [util.types.isDate](https://nodejs.org/api/util.html#utiltypesisdatevalue) should be ok but not [util.isDate](https://nodejs.org/api/deprecations.html#dep0047-utilisdate). Can you confirm if warning is from same file, `lib/connection.js`

sudarshan12s:
I think the warning is coming from [node-oracledb 5.5](https://github.com/oracle/node-oracledb/blob/fa0920010e5ba2a22a5e2c57aee0543ea131c01e/lib/connection.js#L538) 
[util.types.isDate](https://nodejs.org/api/util.html#utiltypesisdatevalue) should also be safe ..  Update to latest node-oracledb should resolve the issue.

sosoba:
Try polyfl:
```js
util.isDate ??= util.types.isDate;
```

sharadraju:
@denizyts @barisaydinayy What is the node-oracledb version that you are using?

denizyts:
> What is the node-oracledb version that you are using?

This deprecation related to oracledb version missmatch on local somehow, my version is 6.10.0 but the connection.js file still same with release 6.3.0 (probably caused by some conflicts with respect to local changes.)  after 6.5.0 "util.types.isDate(value)" is used.  _**So deprecation is gone.**_ But still using "instanceof" is better approach i think.  

Thanks to @sharadraju and @sudarshan12s for responding <3

sharadraju:
Thanks @denizyts. My guess is that your `@types/oracledb` version might still be 6.3. You might want to upgrade that as well.

But we will discuss internally and see if we want to go ahead with `instanceof` in the new version.

PR #1778, opened 2026-07-02
Update package.json to specify node < 22 (which removes the `util.isD…

loucadufault:
These functions are used by oracledb 5, and are dropped in node 22. This marks the compatibility in the `package.json`, which gives a warning. Otherwise an innocent-looking Node upgrade can cause runtime issues which is not fun.

I believe this would need to be merged into the v5.5.x mainline branch in order to target an older release major, but not sure how to accomplish that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "buildpackage": "node package/buildpackage.js"
   },
   "engines": {
-    "node": ">=10.16"
+    "node": ">=10.16 <22"
   },
   "maintainers": [
     {


sharadraju:
We have already moved to node-oracledb 7.0 where all the deprecated functions are removed. Please upgrade to the latest version. Thanks

loucadufault:
@sharadraju right, but this is for users still on the older version as it makes it clear that it requires an upgrade to the newer 7x version

PR #1781, opened 2026-07-10
thin: support O5LOGON for accounts that only carry a 10G password verifier

### LINKED ISSUE #1780: Support O5LOGON authentication for accounts with only a 10G password verifier in Thin mode
**1. Describe your new request in detail**

In Thin mode, connecting as a database user whose password only carries the
legacy 10G verifier (`AUTH_VFR_DATA` verifier type `0x939`) fails with:

```
NJS-116: password verifier type 0x939 is not supported by node-oracledb in Thin mode
```

Thick mode (via Instant Client / OCI) connects to the very same account without
any problem. It would be great for Thin mode to implement the same O5LOGON
handshake so these accounts also work without Instant Client.

This matters for legacy application schemas that still carry only a 10G
verifier — for example because `SEC_CASE_SENSITIVE_LOGON=FALSE` was used
historically, or the password was never re-hashed after a database upgrade.
Today the only workarounds are to switch to Thick mode or to run
`ALTER USER ... IDENTIFIED BY ...` to regenerate the 11G/12C verifiers, which is
not always possible when the account cannot be modified.

I have implemented this and would like to submit a pull request. The handshake
is the same O5LOGON/AES flow already used for 11G/12C — only the source of the
key material differs: the AES-128 key is the 8-byte binary 10G (DES) verifier
padded to 16 bytes with zeros, and the combo key reuses the existing 12C PBKDF2
path (`updateVerifierData`). It has been verified end-to-end against Oracle
Database 19c; the 11G and 12C code paths are unchanged.

One Node-specific note: single-DES was moved to OpenSSL 3's *legacy* provider,
so the 10G verifier is computed with `crypto.createCipheriv('des-ede3-cbc', ...)`
using the 8-byte key tripled (K1=K2=K3, i.e. single DES) — this stays in the
default provider and needs **no** `--openssl-legacy-provider` flag.

I'm aware the 10G verifier is a legacy, weak (case-insensitive, DES-based) hash.
The change does **not** create or encourage 10G verifiers — it only lets Thin
mode authenticate against accounts that already have one, matching Thick-mode
behavior. If you would prefer it gated behind an opt-in (a connection attribute
or environment flag), I'm happy to adjust. **Would you accept such a PR, and in
what form?**

**2. Give supporting information about tools and operating systems. Give relevant product version numbers**

- node-oracledb: 7.0.0, Thin mode
- Node.js: 22 (OpenSSL 3)
- Oracle Database: 19c
- Account: password with only the 10G verifier (verifier type `0x939`)
- Error today: `NJS-116` in Thin mode; the same account connects in Thick mode
- A working implementation (3-file change) is ready on a branch and validated
  A/B: stock 7.0.0 fails `NJS-116`; the patched build connects and runs
  `SELECT 1 FROM dual` against the 0x939 account.

zilent1:
> Resolves #1780 — https://github.com/oracle/node-oracledb/issues/1780

## Summary

In Thin mode, connecting as a user whose password only has the legacy **10G verifier**
(`AUTH_VFR_DATA` verifier type `0x939`) fails with:

```
NJS-116: password verifier type 0x939 is not supported by node-oracledb in Thin mode
```

Thick mode (via Instant Client / OCI) connects to the very same account without issue. This change
implements the same O5LOGON handshake in the Thin driver so these accounts also work in Thin mode.

## Motivation

Legacy databases (long-lived application schemas) frequently still carry only a 10G verifier, either
because `SEC_CASE_SENSITIVE_LOGON=FALSE` was used historically or because the password was never
re-hashed after an upgrade. Until now the only workarounds were to switch to Thick mode or to run
`ALTER USER ... IDENTIFIED BY ...` to regenerate the 11G/12C verifiers — not always possible when the
DBA cannot touch the account. Supporting it in Thin mode removes the Instant Client dependency for
these users.

## What changed

Three small additions, all confined to the existing verifier machinery:

1. **`lib/thin/protocol/constants.js`** — add `TNS_VERIFIER_TYPE_10G: 0x939`.
2. **`lib/thin/protocol/encryptDecrypt.js`** — add `getVerifier10G(username, password)` (and a small
   `_desCbc10G` helper) that computes the 8-byte DES verifier; add a 10G branch in `updateVerifierData`
   that derives the AES key from that verifier. `updateVerifierData` now takes the numeric
   `verifierType` (instead of the `verifier11G` boolean) so it can distinguish 10G/11G/12C.
3. **`lib/thin/protocol/messages/auth.js`**
   - Accept `0x939` in the verifier-type gate (no longer throw `NJS-116`).
   - Do not add the `AUTH_PBKDF2_SPEEDY_KEY` key/value pair for 10G (only 12C sends it).

## How the 10G O5LOGON works (thin)

The handshake is the **same O5LOGON/AES** flow already implemented for 11G/12C — only the *source of
the key material* differs:

- **Key material:** the AES-128 key is the **8-byte binary 10G verifier**, zero-padded to 16 bytes
  (`passwordHash = Buffer.concat([verifier, Buffer.alloc(8)])`). It is **not** the ASCII-hex
  representation of the verifier. The server derives the identical key from the verifier it has stored.
- **Session key:** `partA = AES-128-CBC-decrypt(AUTH_SESSKEY, passwordHash, iv=0)`; `partB` is the
  random client half; the client `AUTH_SESSKEY = AES-128-CBC-encrypt(partB, passwordHash, iv=0)`.
- **Combo key:** `PBKDF2-HMAC-SHA512(hex(partB[:16] || partA[:16]), AUTH_PBKDF2_CSK_SALT,
  AUTH_PBKDF2_SDER_COUNT, keyLen=16)` — i.e. the exact 12C combine path, only with a 16-byte key length.
- **Password:** `AUTH_PASSWORD = AES-128-CBC(comboKey, random16 || password)` — unchanged from 11G/12C.

Because the combine step is byte-for-byte identical to the existing 12C path, the 10G branch **reuses**
it (the code just selects `keyLen = 16`); no separate combine code is added.

The 10G verifier itself is `DES-CBC(DES-CBC(UPPER(user||password) as UTF-16BE, key=fixed), key=prev)`,
taking the last 8-byte block each round — the classic Oracle 10G hash.

## Node/OpenSSL note

Single-DES was moved to OpenSSL 3's **legacy provider**, so `crypto.createCipheriv('des-cbc', ...)`
throws `ERR_OSSL_EVP_UNSUPPORTED` on modern Node. The verifier is therefore computed with
`des-ede3-cbc` using the 8-byte key **tripled** (K1=K2=K3, which is exactly single DES) — this stays
in the default provider and requires **no** `--openssl-legacy-provider` flag. UTF-16BE (which Node has
no native encoding for) is produced with `Buffer.from(s, 'utf16le')` + `buf.swap16()`.

## Testing

Verified end-to-end against **Oracle Database 19c**, Thin mode, no Instant Client:

- A/B on an account carrying only the 10G verifier: stock node-oracledb 7.0.0 fails with
  `NJS-116: password verifier type 0x939 is not supported`; the patched build connects and runs
  `SELECT 1 FROM dual`.
- The DES verifier computation matches the well-known 10G hash vectors
  (`SYSTEM/MANAGER → D4DF7931AB130E37`, `SCOTT/TIGER → F894844C34402B67`).
- **Non-regression:** an account with a modern (11G/12C) verifier still connects — the change is
  confined to the 10G branch; the 11G and 12C code paths are untouched.

_TODO for the PR: add a test under `test/` gated on an environment variable pointing at a 10G-verifier
account, since the 10G verifier cannot be created on a default modern instance without
`SEC_CASE_SENSITIVE_LOGON=FALSE`._

## Security considerations

The 10G verifier is a legacy, case-insensitive DES-based hash and is considered weak. This change does
**not** create or encourage 10G verifiers; it only lets Thin mode authenticate against accounts that
already have one, matching Thick-mode behavior. Maintainers may prefer to gate it behind an opt-in
(connection attribute / environment flag) — happy to adjust. The recommended long-term fix for affected
accounts remains re-hashing the password (`ALTER USER ... IDENTIFIED BY ...`).

## Compatibility notes

- No change to any 11G/12C code path.
- No new dependency — only Node's built-in `crypto`.

## Contributor checklist

- [ ] OCA signed
- [ ] Rebased on current `main`
- [x] Test added under `test/`
- [x] CHANGELOG/docs updated

### DIFF
--- a/doc/src/release_notes.rst
+++ b/doc/src/release_notes.rst
@@ -13,6 +13,17 @@ that impact both Thin and Thick modes ('Common'), the changes that
 affect Thin mode (the default runtime behavior of node-oracledb from 6.0.0),
 and the changes that affect the optional :ref:`Thick Mode <enablingthick>`.
 
+node-oracledb v7.1.0 (unreleased)
+-----------------------------------------
+
+Thin Mode Changes
++++++++++++++++++
+
+#)  Added support for connecting to accounts whose password only carries the
+    legacy 10G (case-insensitive, DES) password verifier, matching Thick mode.
+    Previously, such connections failed with ``NJS-116``. See
+    :ref:`pwverifier`.
+
 node-oracledb `v7.0.0 <https://github.com/oracle/node-oracledb/compare/v6.10.0...v7.0.0>`__ (2 Jun 2025)
 -----------------------------------------------------------------------------------------------------------
 
--- a/doc/src/user_guide/appendix_a.rst
+++ b/doc/src/user_guide/appendix_a.rst
@@ -559,10 +559,9 @@ versions. The password verifier can be 10G (case-insensitive Oracle password
 verifier), 11G (SHA-1-based password verifier), and 12C (SHA-2-based SHA-512
 password verifier).
 
-Node-oracledb Thin mode supports password verifiers 11G and later.
-Node-oracledb Thick mode supports password verifiers 10G and later. To view
-all the password verifiers configured for the user accounts, use the following
-query:
+Node-oracledb Thin and Thick modes support password verifiers 10G and later.
+To view all the password verifiers configured for the user accounts, use the
+following query:
 
 .. code-block:: sql
 
@@ -570,10 +569,6 @@ query:
 
 The ``PASSWORD_VERSIONS`` column lists all the password verifiers that exist
 for the user.
-
-If you try to connect to any supported Oracle Database with node-oracledb Thin
-mode, but the user account is created only with the 10G password verifier,
-then the connection will fail with the :ref:`njs116` error.
 
 .. _pooldiff:
 
--- a/doc/src/user_guide/troubleshooting.rst
+++ b/doc/src/user_guide/troubleshooting.rst
@@ -152,24 +152,13 @@ NJS-116
     :summary: The first row displays the NJS-116 error message. The second row displays the probable cause for the NJS-116 error. The third row displays the possible solution to resolve the NJS-116 error.
 
     * - Message
-      - ``NJS-116: password verifier type 0x939 is not supported by node-oracledb in Thin mode.``
+      - ``NJS-116: password verifier type <verifier-type> is not supported by node-oracledb in Thin mode.``
     * - Cause
-      - Connecting to Oracle Database with node-oracledb Thin mode failed because your user account was only created with a 10G password verifier. Node-oracledb Thin mode supports password verifiers 11G and later. See :ref:`pwverifier`.
+      - Connecting to Oracle Database with node-oracledb Thin mode failed because your user account uses a password verifier that node-oracledb Thin mode does not support. Node-oracledb Thin and Thick modes support password verifiers 10G and later. See :ref:`pwverifier`.
     * - Action
-      - You can either:
-
-        - Database administrators can verify if your username only uses the 10G password verifier with this query:
-
-          .. code-block:: sql
-
-                SELECT USERNAME FROM DBA_USERS
-                WHERE ( PASSWORD_VERSIONS = '10G '
-                OR PASSWORD_VERSIONS = '10G HTTP ')
-                AND USERNAME <> 'ANONYMOUS';
-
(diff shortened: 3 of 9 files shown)


zilent1:
Closing this for now. I'd like to first discuss the enhancement in #1780 and get maintainer feedback on whether (and in what form) you'd accept it before submitting the implementation. The branch remains available, so I can reopen or submit a fresh PR once there's agreement on the approach. Apologies for the noise.

